PR #24, opened 2020-03-08
Split parent and daemon modules

ppalaga:


ppalaga:
Could you please review @gnodet ?

gnodet:
@ppalaga I think the patch is missing the two scripts `mvnd` and `mvns`.  Could you please add them back in the `bin` repository ?

ppalaga:
> Could you please add them back in the `bin` repository ?

Good catch! I had them there, but `bin/` was ignored. Solved by un-ignoring `bin/` and adding them back. Plz. re-review.

PR #27, opened 2020-05-28
JUnit 5 extension for testing mvnd

ppalaga:
@gnodet this is an idea how testing of `mvnd` could work. Esp. please have a look at the first simplistic `SingleModuleTest`. The main features:

* There is a JUnit 5 extension, that takes care for starting and killing the daemon per test class
* The JUnit 5 extension also injects the Client to the test
* Tests can pass a ClientOutput mock to the injected client and perform any build command
* The assertions on the ClientOutput can happen using `Mockito.verify()`. 

I admit I generally do not like Mocking but I thought in this particular case, it is really helpful for writing concise tests.

Some refactoring in the Client was required so that tests can access the events sent by the server. 

This is not good enough to get merged yet (`mvnd` is probably broken in the current state). I just wanted to collect some feedback as early as possible and also to give you a chance to veto some design decisions before I invest too much time into it :)


ppalaga:
d3eb3e4 removed a leftover class.

gnodet:
No objections at first glance, so feel free to continue in the same direction.  It looks very promising.

ppalaga:
Ready for review, good to merge from my PoV.  I have added some more tests and `mvnd` is not broken anymore.

Could you please activate Github Actions?

ppalaga:
I have enabled the CI on my fork. All is green now, however I saw some intermittent errors several times. Let's see how stable it will be in the future.

PR #28, opened 2020-06-02
Native mvnd client

ppalaga:


ppalaga:
Is this good to merge, @gnodet ?

ppalaga:
a9e46df contains a sketch how native tests could work. 

There are two implementations of Client now: 

* DefaultClient used in functional greybox JVM testing and
* NativeTestClient for true blackbox integration testing. NativeTestClient is just a wrapper around the native exicutable.

Unlike in Quarkus, the JVM test extends the native one. I did it like that because JVM tests can be more rich. Maybe we'll find a better way in the future.

Bonus: there are CI jobs for Linux, Windows and Mac. 
The Linux failure here seems to be intermittent. The same commit passed in my fork's CI run: https://github.com/ppalaga/mvnd/actions/runs/129093627

PR #30, opened 2020-06-17
mvnd --install feature

ppalaga:


ppalaga:
@gnodet could you please review?

I'd try to create the first release after this.

client/src/main/java/org/jboss/fuse/mvnd/client/DefaultClient.java
```diff
@@ -73,14 +81,36 @@ public static void main(String[] argv) throws Exception {
         }
     }
 
+    private static void install(boolean overwrite) {
+        final Properties buildProps = loadBuildProperties();
+        final String version = buildProps.getProperty("version");
+        final String rawZipUri = System.getProperty("mvnd.zip.uri", "https://github.com/ppalaga/mvnd/releases/download/"+ version +"/mvnd-" + version + ".zip");
```
gnodet: Shouldn't this be modified to point to the `mvndaemon/mvnd` repo instead ?

client/src/main/java/org/jboss/fuse/mvnd/client/Installer.java
```diff
@@ -0,0 +1,194 @@
+package org.jboss.fuse.mvnd.client;
```
gnodet, line 1: Missing header ...

client/src/main/java/org/jboss/fuse/mvnd/client/DefaultClient.java
ppalaga: I changed it in one of the subsequent commits https://github.com/mvndaemon/mvnd/pull/30/files#diff-86f08db3461cd372b91b0c70cd8606c5R85

client/src/main/java/org/jboss/fuse/mvnd/client/Installer.java
ppalaga: +1

ppalaga:
4461d67 added the missing license headers.

ppalaga:
The CI passed. I guess this is good to merge.

Issue #33, opened 2020-06-28
Maven mojo change ignored


ppalaga:
An itest reproducing the issue: https://github.com/mvndaemon/mvnd/pull/34

Issue #26, opened 2020-05-06
Cannot build Quarkus after removing ~/.m2/repository/io/quarkus/ or ~/.m2/repository/io/quarkus/gizmo


ppalaga:
Steps to reproduce
```
rm -Rf ~/.m2/repository/io/quarkus/gizmo
git clone https://github.com/quarkusio/quarkus.git
cd quarkus
mvnd clean install -DskipTests -DskipITs -Denforcer.skip -Dquarkus.build.skip -DskipDocs
```

Expected: build success
Actual: ArC Processor and Qute Generator are complaining about missing gizmo during compilation:

```
[[1;31mERROR[m] Failed to execute goal [32morg.apache.maven.plugins:maven-compiler-plugin:3.8.0-jboss-2:compile[m [1m(default-compile)[m on project [36mqute-generator[m: [1;31mCompilation failure[m: Compilation failure: 
[[1;31mERROR[m] /home/ppalaga/orgs/quarkus/quarkus/independent-projects/qute/generator/src/main/java/io/quarkus/qute/generator/ValueResolverGenerator.java:[5,24] package io.quarkus.gizmo does not exist
...
```

```
[[1;31mERROR[m] Failed to execute goal [32morg.apache.maven.plugins:maven-compiler-plugin:3.8.0-jboss-2:compile[m [1m(default-compile)[m on project [36marc-processor[m: [1;31mCompilation failure[m: Compilation failure: 
[[1;31mERROR[m] /home/ppalaga/orgs/quarkus/quarkus/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/AnnotationLiteralProcessor.java:[4,24] package io.quarkus.gizmo does not exist
```

It is very strange that there is no trace of gizmo in the local maven repo under `~/.m2/repository/io/quarkus/gizmo` I'd expect maven to download it. I do not see any download related error in the log.

gnodet:
Yes, this is certainly a caching problem.  I suppose this one is caused by the https://github.com/gnodet/mvnd/blob/master/daemon/src/main/java/org/jboss/fuse/mvnd/plugin/CliPluginRealmCache.java which purpose is to cache the classloaders / classes for each plugin.   I think the current check verifies the input used to build the classloader, but not the files stored in the repository.  I assumed the existing files in the repository to be immutable. 
Actually, it could be related to #25 if using snapshots.

ppalaga:
[...]

PR #34, opened 2020-06-28
Reproducer for #33 Maven mojo change ignored

ppalaga:


ppalaga:
Strange that it passed in Mac

ppalaga:
@gnodet would you mind checking locally what is going on Mac? Is it really passing or the Mac CI is a false positive?

gnodet:
> @gnodet would you mind checking locally what is going on Mac? Is it really passing or the Mac CI is a false positive?

The test succeeds on my Macbook Pro...

ppalaga:
> The test succeeds on my Macbook Pro...

Thanks for checking! That's even stranger.

I have a PoC how to address this, at least on Linux. The idea is to remove all realms from the classworld that contain -SNAPSHOT URLs or which are under the basedir of the currently built project. I am also throwing the plexus container and all its dependents away after each build. Let me polish it a bit and amend this PR.

daemon/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
```diff
@@ -181,1156 +166,1194 @@ public int doMain(CliRequest cliRequest) throws Exception
...
-        else if ( cliRequest.commandLine.hasOption( CLIManager.BATCH_MODE )
-            || cliRequest.commandLine.hasOption( CLIManager.LOG_FILE ) )
-        {
-            MessageUtils.setColorEnabled( false );
+
+        @Override
+        public void close() {
+            eventSpyDispatcher.close();
+            container.dispose();
+            purgeRealms();
```
ppalaga, line 284: This is the main idea of the fix: we do not re-use the container (an all its dependents) across mvnd executions and we remove -SNAPSHOTS and in-source-tree jars from the class world.

The fact that we do not ditch the whole classworld is intended to keep as many classes as possible loaded and warm. We throw away only those class loaders that are known to be mutable. We still assume that the rest of the jars in the local Maven repo is immutable. https://github.com/mvndaemon/mvnd/issues/26 is an example of a situation where it does not hold and we should have some checks in place to prevent it.

I was pondering whether we could keep parts of the plexus container across mvnd executions too (to speedup the repeated builds). It seems quite risky.  I am not sure we would be able to reliably remove only all the stale beans and their dependents. The beans can be linked in uncontrollable ways. I have not (quickly) found a way how to do that.

ppalaga:
Oh, I have a theory why the test is passing on Mac: `BasicFileAttributes.fileKey()` probably works differently on Mac than on Linux and windows. It looks fixing the Plugin Cache implementation would be less intrusive than 51f06c7 https://github.com/mvndaemon/mvnd/blob/master/daemon/src/main/java/org/jboss/fuse/mvnd/plugin/CliPluginRealmCache.java#L166

ppalaga:
On Mac, `BasicFileAttributes.fileKey()` seems to change when the content of the file changes https://stackoverflow.com/questions/20676800/inode-value-changed-when-editing-the-file-on-mac-os-x
That's not the case on Linux, where it is constant (device ID and inode). 
I see reports that `BasicFileAttributes.fileKey()` returns null on Windows https://stackoverflow.com/questions/22548076/java-unique-file-id-filekey-path-filekey-ok-filekey-path-possible#comment34317450_22548076

daemon/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
gnodet: Isn't that supposed to be done by the CliPluginRealmCache already ?

gnodet:
> Oh, I have a theory why the test is passing on Mac: `BasicFileAttributes.fileKey()` probably works differently on Mac than on Linux and windows. It looks fixing the Plugin Cache implementation would be less intrusive than [51f06c7](https://github.com/mvndaemon/mvnd/commit/51f06c7d0d90d38a73880f0761e50dc244f2a6ab) https://github.com/mvndaemon/mvnd/blob/master/daemon/src/main/java/org/jboss/fuse/mvnd/plugin/CliPluginRealmCache.java#L166

Ah, that would explain why it works on Mac, it's because that's the OS I developped the `CliPluginRealmCache` on...

daemon/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
ppalaga: Actually yes, see #37 

BTW, I have not see CliPluginRealmCache to be picked. I had to add an explicit binding https://github.com/mvndaemon/mvnd/pull/37/files#diff-085c748cd75673826a31723d0840b82eR486

ppalaga:
Replaced by #37

PR #38, opened 2020-06-29
Fix #33 Maven mojo change ignored

### LINKED ISSUE #33: Maven mojo change ignored

ppalaga:
Turn 3 using WatchService

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -40,6 +40,13 @@ jobs:
     - name: mvn clean verify
       run: ./mvnw clean verify -Pnative -B -ntp -e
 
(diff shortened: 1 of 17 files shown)


ppalaga:
@gnodet I ended up with a hybrid: using the TimestampedCacheRecord we had earlier for Mac and a stock JDK WatchService for all other platforms.

gmethvin/directory-watcher turned up to be too inflexible for our use case - it does not allow adding and removing watched dirs dynamically. We'd have to build and start it anew on every cache add/removal.

ppalaga:
The CI is passing \o/ Could you please review?

PR #82, opened 2020-10-06
Use a pre-populated maven repository for integration tests to speed them up

gnodet:


ppalaga:
-1 for me. Isolation and reproducibility of the tests is important.

gnodet:
Then, we could provide a pre-populated cache, because each test spends 30s downloading tons of dependencies for standard maven plugin.

ppalaga:
I'd be fine with making a shared local optional/configurable for the local development, so that it is still possible to set up the CI to run with isolated local repos.

gnodet:
@ppalaga You care about isolation, right ? I'm proposing to keep the tests isolated the way they currently are (each uses its own repo), but instead of starting from an empty repo, just extract a tar.gz containing all the maven dependencies, so that the maven instance running in the daemon don't have to download them all.  The tests would still be reproducible and isolated.  I don't think we need to test the fact that maven can download its dependencies.  Furthermore, this prevents the tests to be run with no connectivity...

ppalaga:
> @ppalaga You care about isolation, right ? I'm proposing to keep the tests isolated the way they currently are (each uses its own repo), but instead of starting from an empty repo, just extract a tar.gz containing all the maven dependencies, so that the maven instance running in the daemon don't have to download them all. The tests would still be reproducible and isolated. I don't think we need to test the fact that maven can download its dependencies. Furthermore, this prevents the tests to be run with no connectivity...

Yes, I am paranoid :) I really thought it would be better to test that Maven can download and store artifacts under a multi-threaded load. Yes, we use takari local repo, but anyway.

But OK, let's try to make it as efficient as possible. How about leveraging [mrm-maven-plugin](https://www.mojohaus.org/mrm/mrm-maven-plugin/) ? - that's basically a mocked Central backed by the main build's local maven repo. I'd like that solution more because it is self-maintaining. There would be no need to store the blob in git and no need to update it when plugin versions change in the tests. WDYT?

gnodet:
There's definitely a problem with the mrm-maven-plugin.  I suspect it does not support concurrent requests well...
This PR provides a tar-gz'ed repo, but at least it works.

PR #88, opened 2020-10-09
Fix display

gnodet:


daemon/src/main/java/org/apache/maven/cli/logging/Slf4jLogger.java
```diff
@@ -0,0 +1,182 @@
+/*
```
ppalaga, line 1: Too many license headers in this file. Could you please keep only one original which is further below?

daemon/src/main/java/org/apache/maven/cli/logging/Slf4jLogger.java
```diff
@@ -0,0 +1,182 @@
...
+import org.codehaus.plexus.logging.Logger;
+import org.jboss.fuse.mvnd.logging.smart.ProjectBuildLogAppender;
+import org.slf4j.MDC;
+
+/**
+ * Adapt an SLF4J logger to a Plexus logger, ignoring Plexus logger API parts that are not classical and
+ * probably not really used.
+ *
+ * @author Jason van Zyl
+ * @since  3.1.0
```
ppalaga, line 63:
```suggestion
```

I guess we can remove `@since` here

daemon/src/main/java/org/apache/maven/cli/logging/Slf4jLogger.java
```diff
@@ -0,0 +1,182 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import org.codehaus.plexus.logging.Logger;
+import org.jboss.fuse.mvnd.logging.smart.ProjectBuildLogAppender;
+import org.slf4j.MDC;
+
+/**
+ * Adapt an SLF4J logger to a Plexus logger, ignoring Plexus logger API parts that are not classical and
+ * probably not really used.
```
ppalaga, line 60:
```suggestion
 * probably not really used.
 * <p>
 * Adapted from * Adapted from https://github.com/apache/maven/blob/maven-3.6.3/maven-embedder/src/main/java/org/apache/maven/cli/logging/Slf4jLogger.java

```

daemon/src/main/java/org/apache/maven/cli/logging/Slf4jLoggerManager.java
```diff
@@ -0,0 +1,133 @@
+/*
```
ppalaga, line 1: Same issues as with Slf4jLogger.java: licenses, since and link to original

ppalaga (commented):
Some comments inline

daemon/src/main/java/org/apache/maven/cli/logging/Slf4jLogger.java
gnodet: Not really, I think derivative work is covered by the top license and not licensed to the ASF...

PR #94, opened 2020-10-15
Improvements

gnodet:


ppalaga (approved):

ppalaga:
@gnodet there is a real test failure on Windows. I guess the last commit is not correct

PR #96, opened 2020-10-16
Re-layout the distro so that mvn is not in bin

### LINKED ISSUE #91: Re-layout the distro so that mvn is not in bin
The idea was proposed by @rmannibucau on dev@maven.apache.org

> relayout the distribution: mvnd/bin does not contain mvn but set maven.home to a folder not in path with mvn script.

ppalaga:
Fix #91


ppalaga:
The new invoker test is passing, our `mvn` is hidden when `mvnd` is installed via SDK man. All known problems solved from my PoV.
Any more scenarios we should test come to your mind, @gnodet ?

PR #99, opened 2020-10-16
Partial revert to fix windows integration test

gnodet:


ppalaga:
I hope to have a fix for the Max failure in the following PR

Issue #102, opened 2020-10-18
IllegalStateException: Failed to lock offset 0 of .../daemon/registry.bin within 20.0 seconds


ppalaga:
Happens intermittently on all OSes, e.g. here https://github.com/mvndaemon/mvnd/runs/1266804259?check_suite_focus=true
I also have seen happening it locally when building with `-Dmrm=false`

ppalaga:
I have never seen this when using `mvnd` for real builds. I saw it only when running our integration tests. I wonder how much is that a real issue and how much is it given by the way how we run our integration tests. 

Actually `Failed to lock offset 0 of .../daemon/registry.bin` sounds like the file exists but was truncated? If it is the latter we could e.g. consider having separate registry files for the individual itests.

gnodet:
The current test setup uses a mvnd registry specific to each tests: `[mvnd]/integration-tests/target/mvnd-tests/SingleModuleTest/home/.m2/mvnd/registry/0.0.13-SNAPSHOT/registry.bin`

ppalaga:
Another attempt to avoid this issue: https://github.com/mvndaemon/mvnd/pull/244/commits/b8ed8b19af420e782c4de2bdad0161c5fd26d0d0

ppalaga:
Seems to be fixed by 73dec36

PR #117, opened 2020-10-20
Add jansi native libraries

ppalaga:


gnodet (commented):
I'm not really convinced by this commit, since we loose the fact that projects are identified by the `projectGA` output.

ppalaga:
> we loose the fact that projects are identified by the `projectGA` output.

You mean the logging changes in SmartBuilderImpl? https://github.com/mvndaemon/mvnd/pull/117/commits/bd69997eb4f3c87ca62db642a69c112711db5a60

The log messages stay the same. The rationale is that the `projectGA(project)` method does not need to be called if the log level is lower than debug.

For the jansi native libraries commit, I am not 100% sure this is the right approach. In stock Maven the jansi native libs are in `lib/jansi-native/*` and they are not in jars. We do not include those in our distro and I think we should not as long as we use a different jansi version: 1.17.1 on stock maven vs. 1.8 on our side. Do you have any opinion on that?

The CI failure on Linux is intermittent https://github.com/mvndaemon/mvnd/issues/102
The failure on Mac seems to be real - would you mind having a look?

ppalaga:
I have moved the logging change into a separate PR: https://github.com/mvndaemon/mvnd/pull/134

gnodet:
> > we loose the fact that projects are identified by the `projectGA` output.
> 
> You mean the logging changes in SmartBuilderImpl? [bd69997](https://github.com/mvndaemon/mvnd/commit/bd69997eb4f3c87ca62db642a69c112711db5a60)
> 
> The log messages stay the same. The rationale is that the `projectGA(project)` method does not need to be called if the log level is lower than debug.
> 
> For the jansi native libraries commit, I am not 100% sure this is the right approach. In stock Maven the jansi native libs are in `lib/jansi-native/*` and they are not in jars. We do not include those in our distro and I think we should not as long as we use a different jansi version: 1.17.1 on stock maven vs. 1.8 on our side. Do you have any opinion on that?

AFAIK, the main problem is that Jansi does not work when running the natively compiled mvnd.  I don't really care if the native libraries are extracted in the distribution or embedded at the moment.
I'm actively working on Jansi to support missing architectures and fixing windows problems.

> 
> The CI failure on Linux is intermittent #102
> The failure on Mac seems to be real - would you mind having a look?

ppalaga:
> AFAIK, the main problem is that Jansi does not work when running the natively compiled mvnd. I don't really care if the native libraries are extracted in the distribution or embedded at the moment.
> I'm actively working on Jansi to support missing architectures and fixing windows problems.

OK, let's close this then

PR #121, opened 2020-10-21
Fixup #111 Document the number of utilized cores and use 1 core at least

ppalaga:


client/src/main/java/org/jboss/fuse/mvnd/client/DefaultClient.java
```diff
@@ -234,7 +234,7 @@ public ExecutionResult execute(ClientOutput output, List<String> argv) {
 
     static void setDefaultArgs(List<String> args) {
         if (args.stream().noneMatch(arg -> arg.startsWith("-T") || arg.equals("--threads"))) {
-            int procs = Runtime.getRuntime().availableProcessors() - 1;
+            final int procs = Math.max(Runtime.getRuntime().availableProcessors() - 1, 1);
```
ppalaga, line 237: We want at least one core, even on a single-processor machine.

Issue #122, opened 2020-10-21
Environment variables not propagated to the daemon during subsequent builds


zbendhiba:
Hello, I have 3 different versions of graalvm install on my linux machine (RHEL 8).
while usen mvn, it takes the version in $GRAALVM_HOME.
But when I use mvnd, it uses a different version. Is it an issue, or do I need to configure something? How can I switch to another version of GraalVm ?

FYI, I was compiling the quarkus way!!

gnodet:
The java version used to build is determined by the `JAVA_HOME` environment variable.
You can see which one is used by running `mvnd -v` which should output something like:
```
Maven Daemon 0.0.8-darwin-amd64 (native)
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /Users/gnodet/.sdkman/candidates/mvnd/current/mvn
Java version: 11.0.8, vendor: AdoptOpenJDK, runtime: /Users/gnodet/.sdkman/candidates/java/11.0.8.hs-adpt
Default locale: en_GB, platform encoding: UTF-8
OS name: "mac os x", version: "10.15.7", arch: "x86_64", family: "Mac"
```

zbendhiba:
Ok setting GRAALVM_HOME on JAVA_HOME works fine !!
Thanks

ppalaga:
I think this is a general problem with environment variables that we should document at least. The daemon is inheriting the env from the first client that starts it. The env of any subsequent `mvnd` client executions is ignored and that's something we cannot fix in any way. (Is that right, @gnodet?) 

`JAVA_HOME` is a bit exceptional, because the client takes care to choose a matching daemon and if none is available, it starts a new one.

While the following does not work:
```
export GRAALVM_HOME=/some/path
mvnd ... # initial invocation starts a daemon with GRAALVM_HOME=/some/path
export GRAALVM_HOME=/other/path
mvnd .... # the running daemon still sees GRAALVM_HOME=/some/path
```

But the following should work

```
export GRAALVM_HOME=/some/path
mvnd ... # initial invocation starts a daemon with GRAALVM_HOME=/some/path
export GRAALVM_HOME=/other/path
mvnd --stop # kill all running daemons
mvnd ... # starts a fresh daemon with GRAALVM_HOME=/other/path
```

famod:
I wonder whether the daemon (or client?) should compare the current environment to the initial one and issue a warning in case there are differences? Or even fail?

[...]

PR #125, opened 2020-10-22
Display warning in case of environment mismatch #122

gnodet:


daemon/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
```diff
@@ -472,6 +477,27 @@ private void repository(CliRequest cliRequest)
...
+                .collect(Collectors.toList());
+        if (!diffs.isEmpty()) {
+            slf4jLogger.warn("Environment mistmach !");
+            slf4jLogger.warn("A few environment mismatches have been detected between the client and the daemon.");
+            diffs.forEach(key -> {
+                String vr = requested.get(key);
+                String va = actual.get(key);
+                slf4jLogger.warn("   {} -> {} instead of {}", key, va, vr);
+            });
+            slf4jLogger.warn("This is usually harmless.");
```
ppalaga, line 497:
```suggestion
            slf4jLogger.warn("If the difference matters to you, stop the running daemons using mvnd --stop and start a new daemon from the current environment by issuing any mvnd build command");
```

ppalaga (commented):
A suggestion inline. Otherwise looks good.

PR #140, opened 2020-10-24
Do not add mvnd-client.jar to daemon's class path

ppalaga:


gnodet:
I think we may want to remove the mvnd.sh script (unless we create the corresponding `mvnd.bat` windows equivalent) and maybe not include the mvnd-client jar in the distribution at all.  Thoughts ?

ppalaga:
I do not mind keeping `mvnd.sh`. It might come in handy on unix-like platforms for which we do not provide a native executable. Windows is kinda supported (although only 64bit) by a native executable - so I would not see having `mvnd.cmd` for Windows as that urgent.

PR #141, opened 2020-10-24
Remove the superfluous Serializer interface and its implemetation

ppalaga:


ppalaga:
The original MessageSerializer was using the same byte array for both read and write ops. Those reads and writes happen in different threads. To avoid potential conflicts between the read and write threads I have split the MessageSerializer into MessageReader and MessageWriter, each holding its own byte array buffer.

The issue is actually not critical at the moment, because the ops do not happen to interleave: client sends one message and then only receives without sending. This might perhaps change in the future, when we start supporting build cancellation. Anyway, I find the split conceptually cleaner.

That was about the first two commits

Now about the third commit, that I'd like to hear your opinion about:

Having done the split, I am struggling about the necessity of the byteBuf in the [reader](https://github.com/mvndaemon/mvnd/pull/141/files#diff-07780342186ffa696b0e9d532cb9e9cf11f62cae11d2980e44c7be22c0619892R199) and [writer](https://github.com/mvndaemon/mvnd/pull/141/files#diff-07780342186ffa696b0e9d532cb9e9cf11f62cae11d2980e44c7be22c0619892R345) 

The byteBufs do not seem to make much sense performance-wise, because the underlying Data[Input|Output]Streams are are backed by custom Socket[Output|Input]Streams https://github.com/mvndaemon/mvnd/blob/ba56664e2cecafdb77981687b507bb82e57dd7ac/common/src/main/java/org/jboss/fuse/mvnd/common/DaemonConnection.java#L66-L67 
that have a ByteBuffer https://github.com/mvndaemon/mvnd/blob/ba56664e2cecafdb77981687b507bb82e57dd7ac/common/src/main/java/org/jboss/fuse/mvnd/common/DaemonConnection.java#L159

So why could we not further simplify MessageReader and MessageWriter by removing the byteBuf and the [read|write]UTF methods and use java.io.Data[Inpiut|Output]Stream.[read|write]UTF(String) instead?
We'd end up with less code and in theory also a bit better perf. WDYT?

gnodet:
It seems this PR goes back and forth (adding classes + removing those).
At the end, it seems to be about removing the Serializer instance and make the code into static methods and removing the read/writeUTF methods to delegate to the ones from `DataInput/OutputStream`.
That sounds good to me, but we may want to simplify the commits a bit.

ppalaga:
Thanks, squashed.

PR #145, opened 2020-10-25
Mention Homebrew tap in the README, show asciinema cast instead of a …

ppalaga:
…static image


gnodet:
@ppalaga did you create an account on asciinema so that the download never expires ? It happened once to me that the link was invalidated after a year or so...

ppalaga:
Yes, I did

PR #167, opened 2020-10-28
Scanned some class with SonarQube and make some changes to quality of code

juanpereira2:


client/src/main/java/org/jboss/fuse/mvnd/client/DefaultClient.java
```diff
@@ -179,7 +179,7 @@ public ExecutionResult execute(ClientOutput output, List<String> argv) {
                         try {
                             ProcessHandle.of(di.getPid()).ifPresent(ProcessHandle::destroyForcibly);
                         } catch (Exception t) {
-                            System.out.println("Daemon " + di.getUid() + ": " + t);
+                            LOGGER.info("Daemon " + di.getUid() + ": " + t);
```
ppalaga, line 182: I think we want the System.out.println() here. We prefer the message NOT to be prefixed with date, logger name, in this situation where we are killing daemons and some of those kills may fail.

common/src/test/java/org/jboss/fuse/mvnd/common/DaemonRegistryTest.java
```diff
@@ -26,10 +26,10 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
-public class DaemonRegistryTest {
+ class DaemonRegistryTest {
```
ppalaga, line 29: Why does this matter?

BTW this line and the lines below break the indentation.

ppalaga:
Please feel free to reopen if you'd like to explain why this change matters.

PR #181, opened 2020-11-02
Issue 114

gnodet:


common/src/main/java/org/jboss/fuse/mvnd/common/Message.java
```diff
@@ -267,7 +267,7 @@ public static Message read(DataInputStream input) throws IOException {
             String message = readUTF(input);
             String className = readUTF(input);
             String stackTrace = readUTF(input);
-            return new BuildException(message, className, stackTrace);
+            return new BuildException(message.isEmpty() ? null : message, className, stackTrace);
```
ppalaga, line 270: Wouldn't `readUTF()` and `writeUTF()` be a better place for this fix?

gnodet: > Wouldn't `readUTF()` and `writeUTF()` be a better place for this fix?

I thought about it but I rejected it, because this would mean that an empty string will be deserialised to a `null` which could be unexpected.  The current fix is less error-prone imho.

gnodet: @ppalaga actually, it could be done if we extend to support both empty and null strings.  Maybe using a length of -1, which should not be a problem since the length is now an `int`.

ppalaga: > support both empty and null strings. Maybe using a length of -1

Sounds good

Issue #180, opened 2020-11-02
Support for interactive sessions


gnodet:
If a plugin asks the user for anything, we need to be able to carry that on to the `mvnd` client.
For example the `maven-release-plugin` uses `org.codehaus.plexus.components.interactivity.Prompter`, not sure how other plugins do it though.

PR #187, opened 2020-11-04
Support for interactive sessions #180

gnodet:


gnodet:
This PR has a problem : the custom `Prompter` registered at https://github.com/gnodet/mvnd/blob/c97aeea42628b6f457805a04f6047c0568a63647/daemon/src/main/java/org/apache/maven/cli/DaemonMavenCli.java#L476-L478 is *not always* used by plugins.  It also needs an addition integration test.

gnodet:
A possible improvement would be to use JLine's `LineReader` to actually read the prompt...

PR #198, opened 2020-11-09
Client: have just one event queue and one consuming thread 

### LINKED ISSUE #133: Client: have just one event queue and one consuming thread
After https://github.com/mvndaemon/mvnd/pull/123, we have two chained event queues and as many as four threads on the client side:

* main
* TerminalOutput.displayLoop
* TerminalOutput.readInputLoop
* DaemonClientConnection.doReceive()

I'd like to investigate if we could have just a single event queue and one less client thread.

ppalaga:
Fix #133


gnodet:
The final output has a remaining few lines that should be avoided, see https://github.com/gnodet/mvnd/commit/375c3bdb943d80adb4e366923f399fd437ee920b
Otherwise looks good, nice work !

ppalaga:
Thanks, I have cherry-picked 375c3bdb943d80adb4e366923f399fd437ee920b

PR #195, opened 2020-11-09
Implement build cancelation, fixes #127

gnodet:

Issue #193, opened 2020-11-09
The daemon created by the client should survive if the client is interrupted using Ctrl+C


gnodet:
The JVM has a bad habit (for our use case) to still _own_ child processes.
This means that when the JVM receives an interrupt signal, it will forward it to child processes. 
The means that if the client creates a daemon receives a SIGINT, the daemon won't be able to survive.
The daemon needs to call `setsid` in order to detach.  The jansi library would be a good candidate for this function since it already contains a bunch of posix functions.

ppalaga:
Is this the case only when the daemon is started from `mvnd.sh` - i.e. from a stock JVM? I am asking because the daemon seems to survive when I interrupt the `mvnd` native client with `CTRL+C`.

ppalaga:
> the daemon seems to survive when I interrupt the `mvnd` native client with `CTRL+C`.

Plz. forget what I said. The daemon is killed indeed.

gnodet:
Actually, a more simple way may be to simply have the daemon ignore the `INT` and `TSTP` signals, it would avoid any additional JNI requirement.

PR #199, opened 2020-11-10
Implement build cancelation

### LINKED ISSUE #127: Implement build cancellation
Some support is already provided (such as a specific state and such) but we need to intercept the Ctrl+C / interrupt event, send a cancellation request and check on the daemon side.

ppalaga:
fixes #127

### DIFF
--- a/client/src/main/java/org/jboss/fuse/mvnd/client/DaemonClientConnection.java
+++ b/client/src/main/java/org/jboss/fuse/mvnd/client/DaemonClientConnection.java
@@ -32,7 +32,6 @@ import org.jboss.fuse.mvnd.common.DaemonException.ConnectException;
 import org.jboss.fuse.mvnd.common.DaemonException.StaleAddressException;
 import org.jboss.fuse.mvnd.common.DaemonInfo;
 import org.jboss.fuse.mvnd.common.Message;
-import org.jboss.fuse.mvnd.common.Message.Prompt;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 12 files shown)


ppalaga:
This is a bit different than the original https://github.com/mvndaemon/mvnd/pull/195 : the cancellation request is not enforced while the DaemonConnector is connecting. It is rather stored in a temporary outbound queue, waiting for the connection and is send to the daemon only after when the connection is established. Could you please review @gnodet ?

gnodet:
I'd like to keep the ability to cancel the client during the first few seconds while it's trying to connect to a daemon.
I think it currently somewhat works because of https://github.com/mvndaemon/mvnd/issues/193, so I think the daemon is killed very early when hitting `Ctrl+C`.

Also, it seems the client is waiting for the daemon to process the cancellation request cleanly.  I don't think this is a desirable behaviour.  It can slow the response on the client side a few seconds.  The client should send the cancel message and exit early.

gnodet:
> I'd like to keep the ability to cancel the client during the first few seconds while it's trying to connect to a daemon.
> I think it currently somewhat works because of #193, so I think the daemon is killed very early when hitting `Ctrl+C`.
> 
> Also, it seems the client is waiting for the daemon to process the cancellation request cleanly. I don't think this is a desirable behaviour. It can slow the response on the client side a few seconds. The client should send the cancel message and exit early.

I should have a proposal in a few minutes...

gnodet:
@ppalaga see https://github.com/gnodet/mvnd/commit/3a38265fac9bb8ce1695b84f5f3f90c770d4d7db

ppalaga:
Thanks, amended with minor changes in c52dfa3

Issue #131, opened 2020-10-23
Better support for dumb terminals


jglick:
Split from #116. Emacs 26.3. `M-x shell` produces a quasi-shell with

```console
$ echo $TERM
emacs
```

ANSI highlighting is supported, as well as _simple_ cursor movement (so for example dynamic display from `git pull` works fine). `mvnd`’s dynamic display gets mangled, though; some attempt to move between lines is just too much:

```console
$ mvnd -f …/jenkinsci/build-token-root-plugin clean verify
Building...
[...]

PR #204, opened 2020-11-11
Improve display with an easy opt-out option and support for dumb term…

### LINKED ISSUE #116: Option to default to a simple log when using a single thread
It would come in handy to able to opt-out of the log buffering in case you want a more default maven output, e.g. to compare runs to `mvn` in case `mvnd` is showing unexpected behaviour.
When building Quarkus with its 700+ modules, console can also become unresponsive for a while when everything is flushed at once at the end of the build.

gnodet:
…inals, fixes #116 and #131


common/src/main/java/org/jboss/fuse/mvnd/common/Environment.java
```diff
@@ -45,8 +54,16 @@ public String asCommandLineProperty(String value) {
...
+    //
+    // mvnd properties
+    //
     MVND_PROPERTIES_PATH("mvnd.properties.path", "MVND_PROPERTIES_PATH", null, false),
     MVND_DAEMON_STORAGE("mvnd.daemon.storage", null, null, false),
+    /**
+     * Property that can be set to avoid buffering the output and display events continuously, closer to the usual maven
+     * display.
+     */
+    MVND_NO_BUFERING("mvnd.noBuffering", null, "false", false),
```
jglick, line 66: typo, should be `MVND_NO_BUFFERING`

PR #206, opened 2020-11-11
Property to configure the default rolling window size (defaults to 2)

gnodet:


ppalaga:
Ups, looks like this should have come before https://github.com/mvndaemon/mvnd/pull/208 let me try to rebase

ppalaga:
merged as c8c5d08e1f562cf08cfcd8f0b02794f1fcac019c

PR #207, opened 2020-11-11
Do not assume TerminalOutput is updated only on the main thread

ppalaga:


ppalaga:
https://github.com/mvndaemon/mvnd/pull/205 looks better

PR #213, opened 2020-11-12
Automatic purge of daemon logs

gnodet:


common/src/main/java/org/mvndaemon/mvnd/common/TimeUtils.java
```diff
@@ -0,0 +1,244 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Time utils.
+ *
+ * Origin file:
+ * https://github.com/apache/camel/blob/4ea9e6c357371682b855d2d79655b41120331b7a/core/camel-util/src/main/java/org/apache/camel/util/TimeUtils.java
+ */
+public final class TimeUtils {
```
ppalaga, line 34: TimeUtils brings quite a lot of code. I wonder whether I cannot be simplified or perhaps fully replaced by `java.time.Duration.parse(CharSequence)`/`java.time.Duration.toString()`?

common/src/main/java/org/mvndaemon/mvnd/common/Environment.java
```diff
@@ -68,14 +68,18 @@ public String asCommandLineProperty(String value) {
      * The number of log lines to display for each Maven module that is built in parallel.
      */
     MVND_ROLLING_WINDOW_SIZE("mvnd.rollingWindowSize", null, "0", false),
+    /**
+     * The automatic log purge period
```
ppalaga, line 72: A short description of the supported syntax and a couple of duration examples would be nice here and on other Duration options. 

We should slowly start adding `@since` so that users see how old is which option.

common/src/main/java/org/mvndaemon/mvnd/common/TimeUtils.java
gnodet: I'm all for a tiny `mvnd` executable, but I don't think our code size will become a problem at the moment.  When the executable has been trimmed (I'll raise an issue with some infos), we could revisit if needed.

The goal is to be human readable.  The `java.time.Duration.parse`/`toString` methods use the ISO format which looks like `PT8H6M12.345S`, and this is not friendly at all imho.

common/src/main/java/org/mvndaemon/mvnd/common/Environment.java
gnodet: Makes sense, I'll add one.
I think we can start using `@since` when we'll switch to a non `0.0.x` version, which kinda assumes, very early code and non stabilised api.

common/src/main/java/org/mvndaemon/mvnd/common/TimeUtils.java
gnodet: I think the code could be simplified by using a single regex instead of multiple ones.  I'll have a look.

gnodet: I've double checked and the `mvnd` executable already contains the `java.time` packages, mainly because the `java.lang.ProcessHandleImpl$Info.toString()` references it indirectly (in `#totalCpuDuration()`).  I'll try removing this reference to see how much we gain in the final size.

ppalaga: > I think the code could be simplified by using a single regex instead of multiple ones.

That would be nice. I even would not mind supporting only the shortest variants `1d` vs. `1 day`, `1 days`.

ppalaga: > I've double checked and the `mvnd` executable already contains the `java.time` packages, mainly because the `java.lang.ProcessHandleImpl$Info.toString()` references it indirectly (in `#totalCpuDuration()`). I'll try removing this reference to see how much we gain in the final size.

Interesting excercise :)

gnodet: It's not really easy to get rid of the `java.time.*` packages as those are referenced by various classes in the JDK such as `java.util.Formatter`, `java.util.logging.LogRecord`...

common/src/main/java/org/mvndaemon/mvnd/common/TimeUtils.java
```diff
@@ -59,186 +55,67 @@ public static String printDuration(Duration uptime) {
...
+            String d = matcher.group("d");
+            String h = matcher.group("h");
+            String m = matcher.group("m");
+            String s = matcher.group("s");
+            String l = matcher.group("l");
+            return (d != null ? TimeUnit.DAYS.toMillis(Long.parseLong(d)) : 0)
+                + (h != null ? TimeUnit.HOURS.toMillis(Long.parseLong(h)) : 0)
+                + (m != null ? TimeUnit.MINUTES.toMillis(Long.parseLong(m)) : 0)
+                + (s != null ? TimeUnit.SECONDS.toMillis(Long.parseLong(s)) : 0)
+                + (l != null ? TimeUnit.MILLISECONDS.toMillis(Long.parseLong(l)) : 0);
```
ppalaga, line 117: Much cleaner, thanks!

ppalaga (approved):

PR #214, opened 2020-11-13
Ignore INT and TSTP signals in the daemon

gnodet:


ppalaga:
What is the motivation for this change? Is this a way to avoid killing the daemon when the client is interrupted?

ppalaga:
In any case, a comment in Server() would be nice.

PR #217, opened 2020-11-13
Messages associated with projectId from the previous build in the ter…

### LINKED ISSUE #216: Messages associated with projectId from the previous build in the terminal and daemon log
Steps to reproduce:

```
mvnd --stop

cd project-1
mvnd clean verify
# ... the output is correct

cd ../../independent-source-tree/module
mvnd clean verify
# NOT OK: the first line of the output starts with :project-1 - this can also be seen in the server log.
# Expected: no stale projectIds in the output
```

ppalaga:
…minal and daemon log 

Fix #216

I am not 100% sure this is the best fix, WDYT, @gnodet ?

### DIFF
--- a/common/src/main/java/org/mvndaemon/mvnd/common/Environment.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/Environment.java
@@ -77,7 +77,7 @@ public enum Environment {
     /**
      * The automatic log purge period
      */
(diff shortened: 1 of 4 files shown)


gnodet:
> …minal and daemon log
> 
> Fix #216
> 
> I am not 100% sure this is the best fix, WDYT, @gnodet ?

Not sure if that would be sufficient, but I don't really see any drawback in adding that.

Issue #218, opened 2020-11-13
pom.xml changes not honored (post 0.0.10 regression)


ppalaga:
Happens again, probably caused by 67d5b4b9c95e1f56d1287a2efccd33c67b091bdb

Steps to reproduce:

```
mvnd --stop # kill all daemons

git clone -o upstream git@github.com:apache/camel-quarkus.git
cd camel-quarkus

git reset --hard d691671ede042e7837a36efb509f87fa85895f9c
# baseline build
mvnd clean install -Dquickly

# now lets change the poms
git fetch upstream pull/2008/head:pr2008
git checkout pr2008

mvnd clean install -Dquickly
# NOT OK: fails with some version not specified errors
# EXPECTED: should pass normally
```

Reproducible with mvnd 0.0.11, 0.0.12 and current master
Works with mvnd 0.0.10

The bottom line is that I do not get what is different here and in the bom intergration test that is passing daily on our CI.

ppalaga:
I have posted a reproducer: https://github.com/mvndaemon/mvnd/pull/219

ppalaga:
> I do not get what is different here and in the bom intergration test that is passing daily on our CI.

After fixing https://github.com/mvndaemon/mvnd/issues/220 , the `UpgradesInBom*` tests are failing as expected too.

ppalaga:
The reproducers are in the master and they are temporarily set to expect the wrong outcome https://github.com/mvndaemon/mvnd/blob/1ac48446deacf4762650ed0f1ed76eef56cd838e/integration-tests/src/test/java/org/mvndaemon/mvnd/it/NewManagedModuleNativeIT.java#L86 https://github.com/mvndaemon/mvnd/blob/1ac48446deacf4762650ed0f1ed76eef56cd838e/integration-tests/src/test/java/org/mvndaemon/mvnd/it/UpgradesInBomNativeIT.java#L80

PR #219, opened 2020-11-13
Reproduce #218 pom.xml changes not honored

ppalaga:


ppalaga:
Replaced by https://github.com/mvndaemon/mvnd/pull/222

PR #223, opened 2020-11-16
Clean the names of properties, fixes #221

### LINKED ISSUE #221: Clean the names of properties
The environment properties are not always named in a consistent way.
We should make them more coherent.
One solution would be to have all properties beginning with `mvnd.` instead of `daemon.`
Also we have a few properties that use `.` to separate words instead of using a capital such as `mvnd.daemon.storage` or `daemon.core.extensions`.

gnodet:


ppalaga (approved):
I appreciate the comments!

ppalaga (approved):

PR #252, opened 2020-12-08
Use client terminal width to format help

gnodet:


ppalaga (approved):
Looks good, but would you mind merging https://github.com/mvndaemon/mvnd/pull/250 first that is a bit older and has some potential to conflict?

PR #253, opened 2020-12-09
Maven like behaviour and other small improvements

gnodet:


client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
```diff
@@ -58,41 +59,50 @@
...
-                }
-                args.add(arg);
+                continue;
+            }
+            /* This needs to be done very early, otherwise various DaemonParameters do not work properly */
+            final int eqPos = val.indexOf('=');
+            if (eqPos >= 0) {
+                System.setProperty(val.substring(2, eqPos), val.substring(eqPos + 1));
+            } else {
+                System.setProperty(val.substring(2), "");
```
ppalaga, line 103: I guess this should be moved under the `if (arg.startsWith("-D")) {` branch above to work correctly.

ppalaga:
Is there any technical (performance maybe?) reason for the `Pass the arguments list instead of the add method only` commit?
From the design point of view, `Consumer<String>` seemed better to me, because the methods would allow for more general usage like that: except for List, they could be used with any Collection and also with StringBuilder, Writer, or anything else. 
It is a detail, your variant will work too, but I am curious about your motivation anyway.

gnodet:
> Is there any technical (performance maybe?) reason for the `Pass the arguments list instead of the add method only` commit?
> From the design point of view, `Consumer<String>` seemed better to me, because the methods would allow for more general usage like that: except for List, they could be used with any Collection and also with StringBuilder, Writer, or anything else.
> It is a detail, your variant will work too, but I am curious about your motivation anyway.

The reason is that I added a new method (https://github.com/mvndaemon/mvnd/pull/253/commits/7d4e5f8fae89c0b50a102c38565d87e3a3a183d7#diff-e65744bdec30df4a8c9692d35f97c28fef1e81af93a54a3d0dbf5f0a8e8abd88R349-R375) to extract an argument from the command line.  At that point, it thought it would be more understandable to have all related methods operating on the command line (`appendAsCommandLineOption`, `hasCommandOption` and `removeCommandLineOption`) to take a `List<String>` instead of having heterogeneous parameters such as `Consumer<String>`, `Supplier<Stream<String>>` and `Iterable<String>`.   Btw, I think we could use a `Collection<String>` which could be better than `List<String>`.

client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
gnodet: I don't think so.  This code is executed for both `-Dxxx=yyy` and `--define xxx=yyy`.
However, there is a duplication of the same code on lines https://github.com/mvndaemon/mvnd/blob/ab48ddd7157a60adf96da144b36897df2fca3d22/client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java#L167-L174 which could be removed.

gnodet:
> > Is there any technical (performance maybe?) reason for the `Pass the arguments list instead of the add method only` commit?
> > From the design point of view, `Consumer<String>` seemed better to me, because the methods would allow for more general usage like that: except for List, they could be used with any Collection and also with StringBuilder, Writer, or anything else.
> > It is a detail, your variant will work too, but I am curious about your motivation anyway.
> 
> The reason is that I added a new method ([7d4e5f8#diff-e65744bdec30df4a8c9692d35f97c28fef1e81af93a54a3d0dbf5f0a8e8abd88R349-R375](https://github.com/mvndaemon/mvnd/commit/7d4e5f8fae89c0b50a102c38565d87e3a3a183d7#diff-e65744bdec30df4a8c9692d35f97c28fef1e81af93a54a3d0dbf5f0a8e8abd88R349-R375)) to extract an argument from the command line. At that point, it thought it would be more understandable to have all related methods operating on the command line (`appendAsCommandLineOption`, `hasCommandOption` and `removeCommandLineOption`) to take a `List<String>` instead of having heterogeneous parameters such as `Consumer<String>`, `Supplier<Stream<String>>` and `Iterable<String>`. Btw, I think we could use a `Collection<String>` which could be better than `List<String>`.

I've added a few commits to make things even cleaner, and remove the usage of hardcoded command line options completely.

client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
ppalaga: > I don't think so. This code is executed for both `-Dxxx=yyy` and `--define xxx=yyy`.

I see, but then, I guess `arg.substring(2...` should rather be `arg.substring(0...`

> However, there is a duplication of the same code on lines
> 
> https://github.com/mvndaemon/mvnd/blob/ab48ddd7157a60adf96da144b36897df2fca3d22/client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java#L167-L174
> 
> which could be removed.

Not sure, how exactly you'd like to remove. Extracting to a method could work, but simply removing one of the two will not. The first occurrence is important for native mode and the second is important for the test client.

client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
```diff
@@ -135,44 +135,29 @@ public DefaultClient(DaemonParameters parameters) {
...
-            default:
-                if (arg.startsWith("-D")) {
-                    final int eqPos = arg.indexOf('=');
-                    if (eqPos >= 0) {
-                        System.setProperty(arg.substring(2, eqPos), arg.substring(eqPos + 1));
-                    } else {
-                        System.setProperty(arg.substring(2), "");
-                    }
-                }
-                args.add(arg);
```
ppalaga, line 154: As I said earlier, this code matters, if you call JvmTestClient.execute() from a test.

client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
gnodet: Right, that's the reason why the tests are failing.  I'll move the code in a method.

ppalaga (approved):

PR #255, opened 2020-12-10
Bash completion 

### LINKED ISSUE #215: Bash completion
It looks like we could leverage https://github.com/juven/maven-bash-completion/blob/master/bash_completion.bash 

We could either copy and adapt or contribute there.

We'd need to add our own commands like `--stop`, `--status`, etc. (I am not sure we should also add all `-Dmvnd*` props that can be passed on the command line.)

Once we have an adapted script somewhere, we need to figure out, how to activate it using sdkman and Homebrew.

ppalaga:
Fix #215

### DIFF
--- a/common/src/main/java/org/mvndaemon/mvnd/common/Environment.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/Environment.java
@@ -20,15 +20,17 @@ import java.io.InputStream;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
-import java.util.List;
+import java.util.LinkedHashMap;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.Set;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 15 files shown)


common/src/main/java/org/mvndaemon/mvnd/common/Environment.java
```diff
@@ -42,16 +44,20 @@
...
+    PURGE(null, null, null, OptionType.VOID, Flags.OPTIONAL, "mvnd:--purge"),
     /** Prints the status of daemon instances registered in the registry specified by <code>mvnd.registry</code> */
-    STATUS(null, null, null, OptionType.VOID, Flags.OPTIONAL, "--status"),
+    STATUS(null, null, null, OptionType.VOID, Flags.OPTIONAL, "mvnd:--status"),
     /** Stop all daemon instances registered in the registry specified by <code>mvnd.registry</code> */
-    STOP(null, null, null, OptionType.VOID, Flags.OPTIONAL, "--stop"),
+    STOP(null, null, null, OptionType.VOID, Flags.OPTIONAL, "mvnd:--stop"),
     /** Use one thread, no log buffering and the default project builder to behave like a standard maven */
-    SERIAL("mvnd.serial", null, Boolean.FALSE, OptionType.BOOLEAN, Flags.OPTIONAL, "--serial"),
+    SERIAL("mvnd.serial", null, Boolean.FALSE, OptionType.BOOLEAN, Flags.OPTIONAL, "mvnd:-1", "mvnd:--serial"),
```
ppalaga, line 60: @gnodet there was no short option for --serial here. I have put `-1` there. I am open to discuss if you do not like it.

gnodet: That's fine.

ppalaga:
I simplified the locking of the registry file in hope to solve #102. Could you please review those changes, @gnodet ?

Issue #264, opened 2020-12-15
Remove mvnd.builder.rule* and mvnd.builder.rules.provider.* features


ppalaga:
By default mvnd build modules of multi-module source tree in parallel. The ordering of the modules is given by their dependencies. However, sometimes, in real world scenarios, modules may require another module to be built before itself, although there is no Maven dependency relationship between them. The properties `mvnd.builder.rules`, `mvnd.builder.rule`, `mvnd.builder.rules.provider.url` and `mvnd.builder.rules.provider.script` were introduced as an additional mean for the users to control the build order. After some time, we found out that "virtual" dependecies like the following can be used to reach the same goal:

```
        <dependency>
            <groupId>org.group</groupId>
            <artifactId>my-dependency</artifactId>
            <version>${project.version}</version>
            <type>pom</type>
            <scope>test</scope>
            <exclusions>
                <exclusion>
                    <groupId>*</groupId>
                    <artifactId>*</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

We have no evidence that `mvnd.builder.rules`, `mvnd.builder.rule`, `mvnd.builder.rules.provider.url` and `mvnd.builder.rules.provider.script` have found any substantial adoption and hence we would like to deprecate them now and remove them in some future release in favor of the "virtual" dependecies.

gnodet:
@ppalaga so how do we plan to replace the dependencies in camel-quarkus:
  https://github.com/apache/camel-quarkus/blob/master/extensions-jvm/debezium-mongodb/integration-test/pom.xml#L40

ppalaga:
There is perhaps a couple of left-overs that we need to replace with virtual deps, that we generally use already https://github.com/apache/camel-quarkus/blob/cb66cca3e978e49cb5f55d2d63d35332a8632b59/catalog/pom.xml#L62-L75

gnodet:
@ppalaga this is still 4000 lines of dependencies even if they are automatically generated...
Did the quarkus team agreed to go the same way with the stock maven build ?
The benefit of the rules is that they could be specified in the `.mvn/mvnd.properties` and be tuned without disturbing the main project like it had to be done with `camel-quarkus`.

ppalaga:
> @ppalaga this is still 4000 lines of dependencies even if they are automatically generated...

[...]

PR #265, opened 2020-12-15
Deprecate mvnd.builder.rule* and mvnd.builder.rules.provider.* features

ppalaga:
Related #264


gnodet (changes requested):
Looks good, though the `1.3` version number looks a bit strange.  That should propably be `0.3`

ppalaga:
> Looks good, though the `1.3` version number looks a bit strange. That should propably be `0.3`

Indeed, thanks for noticing!

PR #271, opened 2020-12-15
Maven-like rolling output when the build happens to be linear

### LINKED ISSUE #269: Maven-like rolling output when the build happens to be linear
Originally discussed in https://github.com/mvndaemon/mvnd/issues/210.

In situations, when the build happens to be linear, because the dependency graph does not allow any concurrency, noBuffering could be activated automatically. (This is trivially true for single module builds)

@gnodet :

> It should be possible to enhance the smart builder to compute the maximum concurrency of a project, so that we could also use it as a maximum for the number of threads to display.

ppalaga:
Fix #269


daemon/src/main/java/org/mvndaemon/mvnd/daemon/Server.java
```diff
@@ -655,8 +656,13 @@ public void fail(Throwable t) throws Exception {
 
         @Override
         protected void onStartSession(MavenSession session) {
-            queue.add(new BuildStarted(getCurrentProject(session).getName(), session.getProjects().size(),
-                    session.getRequest().getDegreeOfConcurrency()));
+
+            final DependencyGraph smartGraph = DependencyGraph.fromMaven(session);
+            session.setProjectDependencyGraph(smartGraph);
```
ppalaga, line 661: @gnodet this is quite a rough way to do it. I was not able to persuade the guice container to prefer my org.apache.maven.graph.GraphBuilder impl. to the default one. So I gave up and did this replacement instead. I would not mind if you veto this and propose something more elegant.

ppalaga:
I am a bit more content with 385d77e:

* Our custom SmartGraphBuilder is properly instantiated via Guice
* The DependencyGraph now has a width field, that returns 1 if the graph is serial and Integer.MAX_VALUE otherwise. That's enough for the current feature. It can be enhanced later to return the proper width for other purposes.

gnodet (commented):
> I am a bit more content with [385d77e](https://github.com/mvndaemon/mvnd/commit/385d77e205d43da63237da400dc2f4a07d3cb69c):
> 
> * Our custom SmartGraphBuilder is properly instantiated via Guice
> * The DependencyGraph now has a width field, that returns 1 if the graph is serial and Integer.MAX_VALUE otherwise. That's enough for the current feature. It can be enhanced later to return the proper width for other purposes.

Would it be possible to keep the DependencyGraph interface independent of maven ? The algorithm I have only depends on the upstream / downstream methods, so having to create MavenProject objects for nothing will make things harder to test.

daemon/src/main/java/org/mvndaemon/mvnd/builder/DependencyGraph.java
```diff
@@ -80,35 +191,56 @@
                 deps.get(0).forEach(p -> downstreams.get(p).addAll(deps.get(1)));
             }
         }
-        return new DependencyGraph<MavenProject>() {
-            @Override
-            public Stream<MavenProject> getDownstreamProjects(MavenProject project) {
-                return downstreams.get(project).stream();
-            }
+        final boolean serial = !projects.stream().anyMatch(p -> downstreams.get(p).size() > 1);
```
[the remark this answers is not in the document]
gnodet: Actually, is there any case where it returns 1 apart when building a single project ? It looks like the parent/child relationship will make this simple algorithm useless.  Am I missing something ?

common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
```diff
@@ -173,6 +173,11 @@ private boolean doAccept(Message entry) {
             this.maxThreads = bs.getMaxThreads();
             final int maxThreadsDigits = (int) (Math.log10(maxThreads) + 1);
             this.threadsFormat = "%" + (maxThreadsDigits * 3 + 2) + "s";
+            if (maxThreads == 1) {
```
gnodet, line 176: Should we use a configuration flag for that ?

daemon/src/main/java/org/mvndaemon/mvnd/builder/DependencyGraph.java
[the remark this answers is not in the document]
ppalaga: Yeah, it is not perfect, but it works in some simple cases at least and for me it is still an important improvement. As I said, a proper width computation can be added later.

common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
ppalaga: You mean some users might not want this auto-rolling? I must say I like it like this. I would not want to add the flag unless somebody explicitly asks for it. Of course, if you personally want this, I'll add it immediately.

daemon/src/main/java/org/mvndaemon/mvnd/builder/DependencyGraph.java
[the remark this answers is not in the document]
gnodet: I can easily port my computation algorithm, provided you keep the `DependencyGraph<K>` interface.

ppalaga:
> Would it be possible to keep the DependencyGraph interface independent of maven ? The algorithm I have only depends on the upstream / downstream methods, so having to create MavenProject objects for nothing will make things harder to test.
> 
> See https://github.com/gnodet/mvnd/blob/dag-width/daemon/src/test/java/org/mvndaemon/mvnd/builder/DagWidthTest.java

Yeah, that's a valid point. Let me try that.

daemon/src/main/java/org/mvndaemon/mvnd/builder/DependencyGraph.java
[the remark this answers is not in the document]
gnodet: I've rebased on master.  The computation can be done using the following lines: https://github.com/gnodet/mvnd/blob/d77d1f777b5bdba4697d8afaadef730813af81ce/daemon/src/main/java/org/mvndaemon/mvnd/builder/SmartBuilder.java#L200-L203

ppalaga:
2708c12 keeps the DependencyGraph interface independent of Maven.

ppalaga:
fe963d5: adapted your DagWidth implementation

daemon/src/test/java/org/mvndaemon/mvnd/builder/DependencyGraphTest.java
```diff
@@ -29,7 +29,7 @@ public void testRules() {
         MavenProject a = newProject("a"), b = newProject("b"), c = newProject("c");
```
gnodet, line 29: It may be better to rename the class altogether, given we're testing the SmartProjectDependencyGraph

daemon/src/main/java/org/mvndaemon/mvnd/builder/SmartGraphBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+        if (result.hasErrors()) {
+            return result;
+        } else {
+            final ProjectDependencyGraph defaultGraph = result.get();
+            final SmartProjectDependencyGraph smartGraph = SmartProjectDependencyGraph.fromMaven(defaultGraph,
+                    SmartProjectDependencyGraph.getRules(defaultGraph.getSortedProjects(), session));
+            return Result.success(smartGraph);
+        }
+    }
+}
```
gnodet, line 57: Overall, I'm a bit skeptic on the benefits of overriding the graph builder to wrap the existing graph by delegating to the standard graph builder, as we have to cast it later to actually access the correct instance.
It may be easier to simply wrap it when needed.  If we need to avoid wrapping it multiple times (not sure that's actually the case), we can always cache it in the session / execution context somewhere.
The problem is that it ties us to maven internals (which I'd rather avoid as much as possible) with no real benefits imho.

ppalaga: > it ties us to maven internals

That's a valid point

> It may be easier to simply wrap it when needed. If we need to avoid wrapping it multiple times (not sure that's actually the case), 

Yes, we do. The graph is needed on two locations at least. 

> we can always cache it in the session / execution context somewhere.

Yeah, storing it somewhere in the session was my first thought. I have not found any general purpose map in the session or in any of its associated classes. Do we have any better option than system properties, ThreadLocal or a static variable?

gnodet: `MavenSession.getExecutionRequest().getData()` or `MavenSession.getRepositorySession().getData()` or `MavenSession.getRepositorySession().getCache()`.

Given we want to avoid stale informations, the former may be the better as it won't outlive the current build.

At the moment, it's only used once.  I suppose it's used in an additional call to compute the width ?

common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
gnodet: I'm fine with the auto-rolling behaviour.  Let's advertise the change in the release notes and see if people want a switch.  I've read users saying they do really like the buffered output a lot, but for linear builds it may be a bit different.

ppalaga:
c3ac2f6: DependencyGraph wraps ProjectDependencyGraph and stores it in session.getRequest().getData()

If this iteration is not good enough, I do not mind postponing this.

gnodet:
> [c3ac2f6](https://github.com/mvndaemon/mvnd/commit/c3ac2f6fab48a281b83d99652149f0483c478948): DependencyGraph wraps ProjectDependencyGraph and stores it in session.getRequest().getData()
> 
> If this iteration is not good enough, I do not mind postponing this.

Ok, let me check asap.

gnodet:
Could you just add the following change https://github.com/gnodet/mvnd/commit/b1d9bd5cd4ce2a25a527d435535ba6a0e3329d2d ?

Otherwise looks good !

ppalaga:
> Could you just add the following change [gnodet@b1d9bd5](https://github.com/gnodet/mvnd/commit/b1d9bd5cd4ce2a25a527d435535ba6a0e3329d2d) ?

Accepted in 1c5e086

PR #285, opened 2020-12-22
Less global and mutable state

ppalaga:
@gnodet this should be a pure refactoring with no changes of behavior. The main intention is to make the code easier to read and easier to reason about.

### DIFF
--- a/daemon/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
+++ b/daemon/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
@@ -138,21 +138,21 @@ public class DaemonMavenCli {
 
     private final ClassWorld classWorld;
 
-    private DefaultPlexusContainer container;
+    private final DefaultPlexusContainer container;
 
-    private EventSpyDispatcher eventSpyDispatcher;
+    private final EventSpyDispatcher eventSpyDispatcher;
 
-    private ModelProcessor modelProcessor;
+    private final ModelProcessor modelProcessor;
 
-    private Maven maven;
+    private final Maven maven;
 
-    private MavenExecutionRequestPopulator executionRequestPopulator;
+    private final MavenExecutionRequestPopulator executionRequestPopulator;
 
-    private ToolchainsBuilder toolchainsBuilder;
+    private final ToolchainsBuilder toolchainsBuilder;
 
-    private DefaultSecDispatcher dispatcher;
+    private final DefaultSecDispatcher dispatcher;
 
-    private Map<String, ConfigurationProcessor> configurationProcessors;
+    private final Map<String, ConfigurationProcessor> configurationProcessors;
 
     public DaemonMavenCli() throws Exception {
         slf4jLoggerFactory = LoggerFactory.getILoggerFactory();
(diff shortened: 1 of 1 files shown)


gnodet:
I'm still hesitant to deviate the `DaemonMavenCli` too much from the `MavenCli`...

ppalaga:
> I'm still hesitant to deviate the `DaemonMavenCli` too much from the `MavenCli`...

I agree that keeping `DaemonMavenCli` as close as possible to `MavenCli` would simplify donating the Daemon to ASF. OTOH, `DaemonMavenCli` is the main part of Maven that we customize. Thus, when we start backporting our changes to Maven, `MavenCli` will have to change anyway. In that situation, I believe, it will be better to approch `MavenCli` with a clean design in mind. So if this PR brings a cleaner design to `DaemonMavenCli` any divergence from `MavenCli` should not be considered an obstacle. WDYT?

ppalaga:
I have removed the second commit for now and I have simplified the first one as much as possible. Could you please re-review?

daemon/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
```diff
@@ -808,7 +822,10 @@ private void logSummary(ExceptionSummary summary, Map<String, String> references
 
     private static final String ANSI_RESET = "\u001B\u005Bm";
 
-    private void configure(CliRequest cliRequest)
+    private static void configure(
+            CliRequest cliRequest,
+            EventSpyDispatcher eventSpyDispatcher,
+            Map<String, ConfigurationProcessor> configurationProcessors)
```
gnodet, line 828: Same as above...
Those methods are private and need to be given fields, so it's better to make them non static imho.

daemon/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
```diff
@@ -488,24 +505,18 @@ protected void configure() {
...
-            Set<String> providedArtifacts) {
+    private static List<CoreExtensionEntry> loadCoreExtensions(
+            List<CoreExtension> extensions,
+            ClassRealm containerRealm,
+            Set<String> providedArtifacts,
+            ClassWorld classWorld,
+            Slf4jLoggerManager plexusLoggerManager,
+            Logger slf4jLogger,
+            ILoggerFactory slf4jLoggerFactory,
+            Function<CliRequest, TransferListener> createTransferListener) {
```
gnodet, line 519: What's the benefit of using a static method and passing all parameters ?

ppalaga: Actually now, that those fields are final, you are right, the method could be non-static without any danger that we inadvertently change the state. Let me change it.

daemon/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
ppalaga: configure is called twice with different params, so it is IMO safer not to allow reading the instance fields.

ppalaga:
2fc3cec: improved what I was able to. Is this good to merge now?

gnodet:
> [2fc3cec](https://github.com/mvndaemon/mvnd/commit/2fc3cec0c20b3676b06c12d355e1f5ab5d0eb5b2): improved what I was able to. Is this good to merge now?

Yes

PR #286, opened 2020-12-22
DAG width wrong for parent runtime deployment triple

ppalaga:
A test case for a bug in DagWidth


ppalaga:
The main idea of the fix is to run the DagWidth algorithm on a transitive reduction of the original graph https://github.com/mvndaemon/mvnd/pull/286/commits/cb751474081013a6fbc7e73b7c553e233cd48d3e#diff-2567c5271a1b768680b79edefa025cd2515042e76bbd938626afb3c0ff817b2eR314-R345

As noob as I am in the Graph theory, I am not able to tell the Graph with algorithm now works correctly in all cases, but at least I was not able to find a counterexample.

gnodet:
Looks good to me.

PR #289, opened 2020-12-25
Right-pad projectIds to improve mojo readability in the threaded view 

### LINKED ISSUE #288: Right-pad projectIds to improve mojo readability in the threaded view
Something like this:
image

A PR follows

ppalaga:
Fix #288

This PR depends on PR https://github.com/mvndaemon/mvnd/pull/285 , but the idea in itself is independent. In case you veto #285 @gnodet , I can rebase this. 

Once again, this is how it looks like: 
image

### DIFF
--- a/client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
+++ b/client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
@@ -231,7 +231,6 @@ public class DefaultClient implements Client {
             try (DaemonClientConnection daemon = connector.connect(output)) {
                 output.setDaemonDispatch(daemon::dispatch);
                 output.setDaemonReceive(daemon::enqueue);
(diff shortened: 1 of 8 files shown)


gnodet:
I like the idea.

ppalaga:
The last iteration is slightly modified: 

* the groupId of the plugin is not shown for any plugin (originally it was shown only for non-`org.apache.maven.plugins` groups)
* mojo name shown now

image

gnodet:
> The last iteration is slightly modified:
> 
> * the groupId of the plugin is not shown for any plugin (originally it was shown only for non-`org.apache.maven.plugins` groups)
> * mojo name shown now
> 
> image

@ppalaga In that case, wouldn't it make sense to remove the plugin version too ? I don't think it's really useful in that context.

ppalaga:
+1, plugin version removed in d4581c1165a6174f05c31b20210f56b01790ff59

PR #297, opened 2021-01-05
Fix logging to use an inheritable thread local, fixes #295

### LINKED ISSUE #295: Lost logs
Hi,

it seems logs are not pseudo real time with some plugins (like junit-platform-maven-plugin:1.1.0:launch or exec:java/exec:exec), is it because the plugin uses a handler to spy logs? Any way to capture stdout/stderr too and add them in the rolling window?

gnodet:


integration-tests/src/test/projects/exec-output/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+                        <goals>
+                            <goal>exec</goal>
+                        </goals>
+                        <configuration>
+                            <useMavenLogger>false</useMavenLogger>
+                            <executable>java</executable>
+                            <longClasspath>true</longClasspath>
+                            <arguments>
+                                <argument>-classpath</argument>
+                                <classpath/>
```
ppalaga, line 93: Is `<classpath/>` really needed?

ppalaga (approved):
Looks good, one question inline

PR #298, opened 2021-01-05
Support launching processes with inherited IO, fixes #241

gnodet:

PR #299, opened 2021-01-05
[DONT MERGE] inheritIO issue

rmannibucau:
PR to show the issue with several plugins using inherit IO fork (and more precisely a way to reproduce it in IT module).

Question is now: do we go with agent option? It can intercept inheritIO and trigger redirect to file instead of poll them, or just use the process streams maybe?


gnodet:
Could you test with https://github.com/mvndaemon/mvnd/pull/298 ?

gnodet:
I've added the integration test to #298

rmannibucau:
@gnodet works perfectly, very impatient to get next release with sdkman now. Thanks a lot!

PR #301, opened 2021-01-06
Eliminate mutable global state in AbstractLoggingSpy

ppalaga:
This is the second commit from https://github.com/mvndaemon/mvnd/pull/285, simplified and without renaming and moving AbstractLoggingSpy so that it is easier to follow what it does. Could you please review, @gnodet ?


gnodet:
We're loosing the ability to configure the appender, but I don't think this is a problem, so I'm ok with the changes.

ppalaga:
> We're loosing the ability to configure the appender, but I don't think this is a problem, so I'm ok with the changes.

+1 I do not think we need it.

PR #302, opened 2021-01-06
Rename AbstractLoggingSpy to ClientDispatcher and move it to org.mvndaemon.mvnd.daemon

ppalaga:
The last part of the refactoring originally done in https://github.com/mvndaemon/mvnd/pull/285


gnodet:
I don't think moving the class to `org.mvndaemon.mvnd.daemon` is a good idea as it leads to a cycle between the package dependencies between `org.mvndaemon.mvnd.daemon` and `org.mvndaemon.mvnd.logging.smart`.

ppalaga:
> it leads to a cycle between the package dependencies between `org.mvndaemon.mvnd.daemon` and `org.mvndaemon.mvnd.logging.smart`.

Good catch, thanks. In faf49e56c1518ebf47fa479ea3fcb28fdfbca3b6 I want back to splitting an abstract base and queue-aware implementation in org.mvndaemon.mvnd.daemon

PR #307, opened 2021-01-06
Speed up bash completion loading by packaging it as a file in the ZIP…

### LINKED ISSUE #296: Speed up bash completion loading by packaging it as a file in the ZIP distribution
Completion setup encourages to add to bashrc:

    source <(mvnd --completion bash)

Since I did it, creating a new shell has some latency I hadn't before.
I wonder if it wouldn't encourage to either dump it once and source it (instead of regenerating it) or (better) reuse maven one if ~/.maven_bash_completion.bash exists to avoid the duplication.

side question: current plugin completion is very jboss oriented, is it possible to add meecrowave, winegrower, tomee, jbake - or at least use a ~/.mvn[d]_completion_plugins_customrc ?

ppalaga:
… distribution 

Fix #296


rmannibucau:
@ppalaga shouldn't the readme and the line "echo 'source <(mvnd --completion bash)' >> ~/.bashrc" be updated to use "source $MVND_HOME/bin/mvnd-bash-completion.bash" too?

ppalaga:
Yes, a PR is welcome!

rmannibucau:
here it is https://github.com/mvndaemon/mvnd/pull/308

PR #308, opened 2021-01-07
Replace mvnd --completion by a plain source now the bash file is in t…

rmannibucau:
…he distrib


ppalaga (approved):
Thanks, @rmannibucau !

PR #309, opened 2021-01-07
Remove mvnd.builder.rule* and mvnd.builder.rules.provider.* features

### LINKED ISSUE #264: Remove mvnd.builder.rule* and mvnd.builder.rules.provider.* features

ppalaga:
Fix #264


gnodet:
Merged with https://github.com/mvndaemon/mvnd/commit/82df39d9842ae64aeb867911a556a274e21b5167

Issue #310, opened 2021-01-07
Speeding up parallel plugins setup


gnodet:
Here's a thread dump showing that an attempt to resolve a plugin will block all other plugins.  This means that while a maven plugin dependencies are being downloaded, no other plugin realm can be obtained.

```
"BuilderThread 5@6626" daemon prio=5 tid=0x3c nid=NA waiting for monitor entry
  java.lang.Thread.State: BLOCKED
	 waiting for BuilderThread 9@6635 to release lock on <0x1a1a> (a org.apache.maven.plugin.internal.DefaultMavenPluginManager)
	  at org.apache.maven.plugin.internal.DefaultMavenPluginManager.getPluginDescriptor(DefaultMavenPluginManager.java:175)
	  at org.apache.maven.plugin.internal.DefaultMavenPluginManager.getMojoDescriptor(DefaultMavenPluginManager.java:286)
	  at org.apache.maven.plugin.DefaultBuildPluginManager.getMojoDescriptor(DefaultBuildPluginManager.java:244)
	  at org.apache.maven.lifecycle.internal.DefaultLifecycleMappingDelegate.calculateLifecycleMappings(DefaultLifecycleMappingDelegate.java:116)
	  at org.apache.maven.lifecycle.internal.DefaultLifecycleExecutionPlanCalculator.calculateLifecycleMappings(DefaultLifecycleExecutionPlanCalculator.java:265)
	  at org.apache.maven.lifecycle.internal.DefaultLifecycleExecutionPlanCalculator.calculateMojoExecutions(DefaultLifecycleExecutionPlanCalculator.java:217)
	  at org.apache.maven.lifecycle.internal.DefaultLifecycleExecutionPlanCalculator.calculateExecutionPlan(DefaultLifecycleExecutionPlanCalculator.java:126)
	  at org.apache.maven.lifecycle.internal.DefaultLifecycleExecutionPlanCalculator.calculateExecutionPlan(DefaultLifecycleExecutionPlanCalculator.java:144)
	  at org.apache.maven.lifecycle.internal.builder.BuilderCommon.resolveBuildPlan(BuilderCommon.java:97)
	  at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:111)
	  at org.mvndaemon.mvnd.builder.SmartBuilderImpl.buildProject(SmartBuilderImpl.java:178)
[...]

PR #311, opened 2021-01-07
Speed up parallel plugin setup #310

gnodet:


daemon/src/main/java/org/mvndaemon/mvnd/plugin/CliPluginDescriptorCache.java
```diff
@@ -0,0 +1,240 @@
...
+import org.codehaus.plexus.component.repository.ComponentDescriptor;
+import org.eclipse.aether.RepositorySystemSession;
+import org.eclipse.aether.repository.LocalRepository;
+import org.eclipse.aether.repository.RemoteRepository;
+import org.eclipse.aether.repository.WorkspaceRepository;
+import org.eclipse.sisu.Priority;
+import org.eclipse.sisu.Typed;
+
+/*
+ * gnodet: This file is based on maven DefaultPluginDescriptorCache and changed in order
```
ppalaga, line 46: It would be handy to have a separate commit with the unchanged original file to see what exactly we have changed.

ppalaga (commented):
Do we have at least one test (which?) covering this?

gnodet:
> Do we have at least one test (which?) covering this?

Not at the moment.  Not really sure how to test that for now.  Once I'm done with adding transfer events back to the client, it may be easier to test that downloads can be done in parallel.

daemon/src/main/java/org/mvndaemon/mvnd/plugin/CliPluginDescriptorCache.java
gnodet: Yes, that makes sense.

ppalaga:
> Once I'm done with adding transfer events back to the client, it may be easier to test that downloads can be done in parallel.

Thanks for explaining, no problem with that plan. Because we have a bunch of unreleased features in the master, how about releasing now and have more time for maturing this?

gnodet:
> Because we have a bunch of unreleased features in the master, how about releasing now and have more time for maturing this?

+1

Issue #284, opened 2020-12-21
The output of ConsoleMavenTransferListener looks ugly in the client


ppalaga:
```
[INFO] [stdout] Downloading from central: https://repo.maven.apache.org/maven2/org/apache/maven/maven-settings/3.0/maven-settings-3.0.jar'}
BuildLogMessage{payload='[INFO] [stdout] Progress (4): 175/472 kB | 52 kB | 16/222 kB | 49/527 kB
Progress (4): 175/472 kB | 52 kB | 20/222 kB | 49/527 kB
Progress (4): 178/472 kB | 52 kB | 20/222 kB | 49/527 kB
Progress (4): 178/472 kB | 52 kB | 25/222 kB | 49/527 kB
Progress (4): 178/472 kB | 52 kB | 29/222 kB | 49/527 kB
Progress (4): 181/472 kB | 52 kB | 29/222 kB | 49/527 kB
Progress (4): 181/472 kB | 52 kB | 33/222 kB | 49/527 kB
Progress (4): 181/472 kB | 52 kB | 37/222 kB | 49/527 kB
Progress (4): 184/472 kB | 52 kB | 37/222 kB | 49/527 kB
Progress (4): 184/472 kB | 52 kB | 41/222 kB | 49/527 kB
Progress (4): 184/472 kB | 52 kB | 41/222 kB | 53/527 kB
Progress (4): 186/472 kB | 52 kB | 41/222 kB | 53/527 kB
Progress (4): 186/472 kB | 52 kB | 45/222 kB | 53/527 kB
...
```

I see two options:

1. Use `Slf4jMavenTransferListener` instead of `ConsoleMavenTransferListener`
2. Have a dedicated set of messages for the transfer progress and have some proper progress monitor UI on the client side. 

Option 1. is much easier than 2. and we could perhaps do 1. as a short term improvement even if we agree to aim at 2. WDYT, @gnodet ?

gnodet:
It sounds like a good plan to me.

gnodet:
The transfer events should be conveyed to the client with a new `Message` for correct display on the client.

PR #313, opened 2021-01-07
Send transfer events to the client for better display, #284

gnodet:
This is a first pass at a better display of download events on the client.
@ppalaga could you give it a try see how it can be improved ?

### DIFF
--- a/daemon/src/main/java/org/mvndaemon/mvnd/daemon/ClientDispatcher.java
+++ b/daemon/src/main/java/org/mvndaemon/mvnd/daemon/ClientDispatcher.java
@@ -114,7 +114,29 @@ public class ClientDispatcher extends BuildEventListener {
     }
 
     public void transfer(String projectId, TransferEvent e) {
-        int event = e.getType().ordinal();
+        int event;
+        switch (e.getType()) {
+            case INITIATED:
+                event = Message.TRANSFER_INITIATED;
+                break;
+            case STARTED:
+                event = Message.TRANSFER_STARTED;
+                break;
+            case PROGRESSED:
+                event = Message.TRANSFER_PROGRESSED;
+                break;
+            case CORRUPTED:
+                event = Message.TRANSFER_CORRUPTED;
+                break;
+            case SUCCEEDED:
+                event = Message.TRANSFER_SUCCEEDED;
+                break;
+            case FAILED:
+                event = Message.TRANSFER_FAILED;
+                break;
+            default:
+                throw new IllegalStateException();
+        }
         int request = e.getRequestType().ordinal();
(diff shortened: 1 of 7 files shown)


ppalaga:
I have not looked into the sources, I just tried it on a single module project with `-Dmaven.repo.local=/tmp/empty`. Because it was single module, mvnd auto-switched to `-Dmvnd.noBuffering` and I actually have not seen any transfer progress in the log - is that expected? I'll experiment more today.

ppalaga:
> I have not looked into the sources, I just tried it on a single module project with `-Dmaven.repo.local=/tmp/empty`. Because it was single module, mvnd auto-switched to `-Dmvnd.noBuffering` and I actually have not seen any transfer progress in the log - is that expected? I'll experiment more today.

I reviewed the code and now I see why it is happening. Perhaps the new `DaemonMavenTransferListener` should do its current stuff but also delegate to `Slf4jMavenTransferListener` so that the transfer messages can also be seen in the text log?

daemon/src/main/java/org/mvndaemon/mvnd/daemon/Server.java
```diff
@@ -567,6 +567,15 @@ int getClassOrder(Message m) {
...
             return 4;
+        case Message.TRANSFER_INITIATED:
+        case Message.TRANSFER_STARTED:
+            return 40;
+        case Message.TRANSFER_PROGRESSED:
+            return 41;
+        case Message.TRANSFER_CORRUPTED:
+        case Message.TRANSFER_SUCCEEDED:
+        case Message.TRANSFER_FAILED:
+            return 42;
```
ppalaga, line 578: At some occasion in the future, we could perhaps change the Message IDs to match this ordering so that we do not need an extra comparator method.

daemon/src/main/java/org/mvndaemon/mvnd/daemon/ClientDispatcher.java
```diff
@@ -112,6 +113,19 @@ public void log(String msg) {
         queue.add(Message.log(trimTrailingEols(msg)));
     }
 
+    public void transfer(String projectId, TransferEvent e) {
+        int event = e.getType().ordinal();
```
ppalaga, line 117: This looks fragile for the long term. Once they add/remove/reorder the members in `EventType`, our code stops working properly, but the errors will bubble up during the deserialization on the client side and they be hard to track back here. I think it would be safer to have some sort of explicit `int[]` (by EventType.ordinal()) or `EnumMap<EventType, Integer>` mapping to Message.type here.

ppalaga (commented):
Added some comments inline. Otherwise looks good!

PR #318, opened 2021-01-08
Display the daemon id and shorten it a bit, fixes #314

### LINKED ISSUE #314: Should print id of daemon that is processing the request
Yesterday, the daemon seemed stuck while builing Quarkus (it was actually downloading artifacts from a very slow source).
Another daemon was present so it was not immediately obvious which one was affected (e.g. for `jstack` or for stopping/killing).

I think it would come handy if the client would print the pid of the daemon that is processing the request.

PS: I could have probably pressed + to see what's going on but nevertheless I do think this would be helpful.

gnodet:


ppalaga (commented):
Yes, this is one possible way to implement it. I wonder if we really need to wait for the Daemon to reveal the Daemon ID? We actually know it in the client before we start/connect to the Daemon. So the round trip is perhaps not necessary?

gnodet:
> Yes, this is one possible way to implement it. I wonder if we really need to wait for the Daemon to reveal the Daemon ID? We actually know it in the client before we start/connect to the Daemon. So the round trip is perhaps not necessary?

Do you mean setting the daemon id directly after the connection ?
https://github.com/mvndaemon/mvnd/blob/aae395e8b2365749347eb7617156df88e225d483/client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java#L232-L233

That's fine with me too.  I agree it's a bit more simple ;-)

ppalaga:
Yes, something like that. I did not have any specific location in mind.

ppalaga (approved):

PR #321, opened 2021-01-08
Fix TestUtils imports

famod:
Got uncommitted files after build, without touching anything. Suppose this is a leftover from https://github.com/mvndaemon/mvnd/commit/0f438bdb09df11fc86be93a5e89ee26dc0b21b8b


famod:
Btw, mabye use https://code.revelc.net/impsort-maven-plugin/check-mojo.html in CI?

ppalaga (approved):

Issue #276, opened 2020-12-17
Compilation fails when using a plugin and try to provide additional dependencies to that plugin


lucamolteni:
See the reproducer here https://github.com/lucamolteni/mvnd-plugin-reproducer

The bootstrap module is a simple maven plugin that generates a class in generated-sources

The startup module uses this plugin and print the constant generated in the class

This fails with `mvnd clean install` but runs with `mvn clean install`

```
mvn clean install
java -cp startup/target/java -cp startup/target/startup-1.0.0-SNAPSHOT.jar org.reproducer.MainClass```
```

By removing this section in startup module

```
<!-- the following is used to provide a SLF4j binding to the ValidationBootstrapMain and related CP. Set src/test/resources/logback.xml root to DEBUG for debugging -->
<includePluginDependencies>true</includePluginDependencies>
<additionalClasspathElements>
    <additionalClasspathElement>${project.basedir}/src/test/resources/</additionalClasspathElement>
</additionalClasspathElements>
```

Everything works

cc @tarilabs

gnodet:
Fwiw, it fails only the second time.

lucamolteni:
@gnodet any news? I'd love to use `mvnd` in Drools 

Thanks for the great product!

gnodet:
> @gnodet any news? I'd love to use `mvnd` in Drools
> 
> Thanks for the great product!

I found a way to make the test actually reproductible. 
However, this looks rather a bug in the maven-compiler-plugin .

Here's the log when the compiler fails:
```
ProjectLogMessage{projectId='startup', message='[DEBUG] Using compiler 'javac'.'}
ProjectLogMessage{projectId='startup', message='[DEBUG] Adding /Users/gnodet/work/git/mvnd/integration-tests/target/mvnd-tests/BootstrapPluginTest/project/startup/target/generated-sources/annotations to compile source roots:
  /Users/gnodet/work/git/mvnd/integration-tests/target/mvnd-tests/BootstrapPluginTest/project/startup/src/main/java
  /Users/gnodet/work/git/mvnd/integration-tests/target/mvnd-tests/BootstrapPluginTest/project/startup/target/generated-sources/bootstrap'}
ProjectLogMessage{projectId='startup', message='[DEBUG] New compile source roots:
  /Users/gnodet/work/git/mvnd/integration-tests/target/mvnd-tests/BootstrapPluginTest/project/startup/src/main/java
[...]

PR #330, opened 2021-01-15
Do not run server threads as daemon as it causes problems with the exec-maven-plugin (#276)

gnodet:


gnodet:
The test was failing just after 0.2.0 has been released.  It can't be reproduced with the current HEAD.

PR #340, opened 2021-01-24
Pre-release polishing

ppalaga:


gnodet:
@ppalaga The WatchService should only be used on Windows. The WatchService on linux and OSX uses a polling mechanism with a default of 10s between polls.  Unix systems can rely on timestamp, there's no reason to use the watch service on linux afaik.

ppalaga:
> @ppalaga The WatchService should only be used on Windows. 

Yes, that's still the case. I have just removed the injection in DefaultCacheFactory, so that we do not create the potentialy costy WatcherService on platforms where it is not needed. 

I have created unit tests for both factories and I plan to disable the cases that do not work. Funny enough the timestamp cache test seems to pass on Windows.

> The WatchService on linux and OSX uses a polling mechanism with a default of 10s between polls. Unix systems can rely on timestamp, there's no reason to use the watch service on linux afaik.

The test on Linux is passing with delay of 100ms between the file change op and poll triggered from WatchServiceCache.
You may well be right with those 10s on Mac. Let me increase the delay for Mac and see what it does. I am just curious how it really behaves, I do not mind disabling the test on non-Windows in the end.

> I disagree with moving merging the cache factories with the caches. I don't mind the moving of the cache factory and implementation to `org.mvndaemon.mvnd.cache.impl`, but the actual usage of the caches should be separated from the cache and only use the cache api.

I do not feel like I did what you imply I did, but at the same time, I do not understand fully, what you mean. So could you please rephrase/give more details? Which of the commits is that?

gnodet:
> > @ppalaga The WatchService should only be used on Windows.
> 
> Yes, that's still the case. I have just removed the injection in DefaultCacheFactory, so that we do not create the potentialy costy WatcherService on platforms where it is not needed.
> 
> I have created unit tests for both factories and I plan to disable the cases that do not work. Funny enough the timestamp cache test seems to pass on Windows.
> 
> > The WatchService on linux and OSX uses a polling mechanism with a default of 10s between polls. Unix systems can rely on timestamp, there's no reason to use the watch service on linux afaik.
> 
> The test on Linux is passing with delay of 100ms between the file change op and poll triggered from WatchServiceCache.
> You may well be right with those 10s on Mac. Let me increase the delay for Mac and see what it does. I am just curious how it really behaves, I do not mind disabling the test on non-Windows in the end.

The problem with windows is that the fileKey is always null, so the tests solely rely on the timestamp, which imho is not sufficient.  It is sufficient, however, on unix based systems I think.
The 10s is when using the default.  I did modify the usage using the specific options https://github.com/mvndaemon/mvnd/blob/a9b69c6dee507af72d4790a14aac5726227a202b/daemon/src/main/java/org/mvndaemon/mvnd/cache/factory/WatchServiceCacheFactory.java#L92-L99.  This should reduce the polling period to a more reasonable 2s I think.  This is completely hidden however, and you need to look at the openjdk implementation to find those values.

> 
> > I disagree with moving merging the cache factories with the caches. I don't mind the moving of the cache factory and implementation to `org.mvndaemon.mvnd.cache.impl`, but the actual usage of the caches should be separated from the cache and only use the cache api.
> 
> I do not feel like I did what you imply I did, but at the same time, I do not understand fully, what you mean. So could you please rephrase/give more details? Which of the commits is that?

The files `Cli*` classes should not be in the same package than the `*CacheFactory` classes
https://github.com/mvndaemon/mvnd/tree/23fb7dcd7a1bc47a828b8e44cc95d1b1b6533404/daemon/src/main/java/org/mvndaemon/mvnd/cache/impl

ppalaga:
> The problem with windows is that the fileKey is always null, so the tests solely rely on the timestamp, which imho is not sufficient. It is sufficient, however, on unix based systems I think.
> The 10s is when using the default. I did modify the usage using the specific options
> 
> https://github.com/mvndaemon/mvnd/blob/a9b69c6dee507af72d4790a14aac5726227a202b/daemon/src/main/java/org/mvndaemon/mvnd/cache/factory/WatchServiceCacheFactory.java#L92-L99
> 
> . This should reduce the polling period to a more reasonable 2s I think. This is completely hidden however, and you need to look at the openjdk implementation to find those values.

I have added a comment to DefaultCacheFactory explaining the motivations.

> > > I disagree with moving merging the cache factories with the caches. I don't mind the moving of the cache factory and implementation to `org.mvndaemon.mvnd.cache.impl`, but the actual usage of the caches should be separated from the cache and only use the cache api.
> > 
> > 
> > I do not feel like I did what you imply I did, but at the same time, I do not understand fully, what you mean. So could you please rephrase/give more details? Which of the commits is that?
> 
> The files `Cli*` classes should not be in the same package than the `*CacheFactory` classes
> https://github.com/mvndaemon/mvnd/tree/23fb7dcd7a1bc47a828b8e44cc95d1b1b6533404/daemon/src/main/java/org/mvndaemon/mvnd/cache/impl

Yes, makes sense. How about moving the factory impls to a new package named `org.mvndaemon.mvnd.cache.factory.impl`?

BTW, I wonder why we put the `Cli` prefix on the class names in `org.mvndaemon.mvnd.cache.impl`? Wouldn't e.g. `Invalidating*` be better instead?

ppalaga:
> How about moving the factory impls to a new package named org.mvndaemon.mvnd.cache.factory.impl?
>
> BTW, I wonder why we put the Cli prefix on the class names in org.mvndaemon.mvnd.cache.impl? Wouldn't e.g. Invalidating* be better instead?

I did both in the last iteration. Good to merge and release from my PoV. Could you please perform the final review?

gnodet:
@ppalaga I have an additional fix and another proposal for the package names.  I don't really like the split between `o.m.mvnd.cache` and `o.m.mvnd.cach.factory.impl`...
https://github.com/ppalaga/mvnd/pull/1

gnodet:
@ppalaga I think you're refactoring lost the added `@Inject` on the constructor at https://github.com/mvndaemon/mvnd/blob/9d7b5f9e20258249cb6ee908dfa5cf76a894fbfb/daemon/src/main/java/org/mvndaemon/mvnd/cache/invalidating/InvalidatingPluginArtifactsCache.java#L55

Issue #234, opened 2020-11-18
Environment mismatches should ignore the PWD var


gnodet:
```
[WARN] Environment mismatch !
[WARN] A few environment mismatches have been detected between the client and the daemon.
[WARN]    PWD -> /Users/gnodet/work/git/maven instead of /Users/gnodet/work/git/plexus-utils
[WARN] If the difference matters to you, stop the running daemons using mvnd --stop and
[WARN] start a new daemon from the current environment by issuing any mvnd build command.
```

gnodet:
Superseded by #352

PR #353, opened 2021-02-10
Make sure the environment is properly updated, fixes #352

### LINKED ISSUE #352: Make sure the environment is properly updated
The current dir and environment variables should be set to the ones of the client, see #125 #234.
This will require 2 native methods from jansi to access the stdlib `chdir` and `setenv` functions.

gnodet:


ppalaga:
These changes introduce potential incompatibilities with existing or future JVM vendors/versions we currently do not test with. The same holds for other operating systems on which we currently do not test. 
Is there any known problem except for #234 why we need this? Is #234 serious enough to justify a fix as risky as this one?

gnodet:
> These changes introduce potential incompatibilities with existing or future JVM vendors/versions we currently do not test with. The same holds for other operating systems on which we currently do not test.
> Is there any known problem except for #234 why we need this? Is #234 serious enough to justify a fix as risky as this one?

I think the fact that the current directory is not updated is serious enough problem, see the integration test which clearly demonstrates this fact.
We can always fix future incompatibles if needed, and if any error occurs, the code falls back to the previous behaviour, i.e. warning the user of the mismatch.  Fwiw, gradle does the same kind of trick.

gnodet:
The build obviously fail until a release with https://github.com/fusesource/jansi/commit/6d61a763ebc4a97399e57c38e7c2e0eac7b8cf3c is provided.

ppalaga:
> gradle does the same kind of trick.

That resolved all my concerns :D

ppalaga (approved):

gnodet:
Merged

Issue #358, opened 2021-02-15
Fix color output for file / tty


gnodet:
Related to https://github.com/mvndaemon/mvnd/issues/355

PR #359, opened 2021-02-15
Fix color output for file / tty #358

gnodet:


ppalaga:
Sorry, I know I am a bit late to the party, I am reviewing all PRs I have missed only now before a release.

I wonder if we really need a mvnd specific option for this? I mean couldn't we mimic the behavior of Maven and its `--batch/-B` option? The auto-detection logic using `CLibrary.isatty(1) == 0` could stay as you did it, but instead of `AttributedString.stripAnsi(s)` we could just ṕass `-B` to the backend. If the user strongly prefers colors off, he can pass `-B` explicitly. I think we do not need to support explicit "colors on" option. Maven does not have it either. Let me try to implement it to check whether it is possible at all.

gnodet:
@ppalaga maven should also go with a `--color` option soon hopefully https://issues.apache.org/jira/browse/MNG-7080?filter=12350130

ppalaga:
> @ppalaga maven should also go with a `--color` option soon hopefully https://issues.apache.org/jira/browse/MNG-7080?filter=12350130
> And the _color on_ option would be interesting when you want colored output to a file.

Yeah, I have found out in between, that stock Maven supports `-Dstyle.color` and indeed, leveraging it brings also a support for explicit "color on". My attempt is in https://github.com/mvndaemon/mvnd/pull/371

PR #367, opened 2021-02-23
Fix display showing more projects than the ones actually active

gnodet:


daemon/src/main/java/org/mvndaemon/mvnd/daemon/Server.java
```diff
@@ -430,7 +430,7 @@ private void cancelNow() {
     private void handle(DaemonConnection connection, BuildRequest buildRequest) {
         updateState(Busy);
         final BlockingQueue<Message> sendQueue = new PriorityBlockingQueue<>(64,
-                Comparator.comparingInt(this::getClassOrder).thenComparingLong(Message::timestamp));
```
ppalaga, line 433: If you really wanted to remove getClassOrder from the comparator, then the getClassOrder() method can be removed. 
And we perhaps also do not need a PriorityBlockingQueue anymore.

common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
```diff
@@ -279,8 +281,8 @@ private boolean doAccept(Message entry) {
             Project prj = projects.remove(artifactId);
             if (prj != null) {
                 prj.log.forEach(log);
+                doneProjects.add(artifactId);
```
ppalaga, line 284: Why do we need `doneProjects` set? Under which circumstances would deliver a different count than the original `int doneProjects`?

PR #371, opened 2021-03-07
Leverage Maven's -Dstyle.color to avoid coloring instead of stripping the ASCII codes in the client

### LINKED ISSUE #373: A daemon started with -B/--batch option stays colorless forever
Once the daemon is started with `-B/--batch` option, all subsequent invocations, albeit without `-B`, are colorless.

ppalaga:
Fix #373

### DIFF
--- a/client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
+++ b/client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
@@ -41,6 +41,7 @@ import org.mvndaemon.mvnd.common.DaemonException;
 import org.mvndaemon.mvnd.common.DaemonInfo;
 import org.mvndaemon.mvnd.common.DaemonRegistry;
 import org.mvndaemon.mvnd.common.Environment;
+import org.mvndaemon.mvnd.common.Environment.Color;
 import org.mvndaemon.mvnd.common.Message;
 import org.mvndaemon.mvnd.common.Message.BuildException;
 import org.mvndaemon.mvnd.common.Message.BuildFinished;
@@ -83,25 +84,16 @@ public class DefaultClient implements Client {
                 || Environment.COMPLETION.hasCommandLineOption(args);
 
         // Color
-        String styleColor = Environment.MAVEN_COLOR.getCommandLineOption(args);
-        if ("always".equals(styleColor) || "never".equals(styleColor)) {
-            /* If the user knows what he wants, pass his preference unchanged to the daemon */
-        } else if ("auto".equals(styleColor)) {
-            /* Do not pass auto to daemon, we handle it below */
-            Environment.MAVEN_COLOR.removeCommandLineOption(args);
-            styleColor = null;
-        } else if (styleColor != null) {
-            throw new IllegalArgumentException("Invalid color configuration option [" + styleColor
-                    + "]. Supported values are (auto|always|never).");
+        Color styleColor = Color.of(Environment.MAVEN_COLOR.removeCommandLineOption(args)).orElse(Color.auto);
+        if (styleColor == Color.auto) {
+            /* Translate from auto to either always or never */
+            /* stdout is not a terminal e.g. when stdout is redirected to a file */
+            final boolean stdoutIsTerminal = CLibrary.isatty(1) != 0;
+            styleColor = (batchMode || logFile != null || !stdoutIsTerminal) ? Color.never : Color.always;
         }
-
-        /* stdout is not a terminal e.g. when stdout is redirected to a file */
-        final boolean stdoutIsTerminal = CLibrary.isatty(1) != 0;
-        if (styleColor == null) {
-            Environment.MAVEN_COLOR.addCommandLineOption(
-                    args,
-                    (batchMode || logFile != null || !stdoutIsTerminal) ? "never" : "always");
-        }
+        /* We cannot use Environment.addCommandLineOption() because that one would pass --color to the daemon
+         * and --color is not supported there yet. */
+        args.add("-D" + Environment.MAVEN_COLOR.getProperty() + "=" + styleColor.name());
 
         // System properties
         setSystemPropertiesFromCommandLine(args);
@@ -114,7 +106,7 @@ public class DefaultClient implements Client {
         }
 
         int exitCode = 0;
-        boolean noBuffering = batchMode || parameters.noBuffering() || !stdoutIsTerminal;
+        boolean noBuffering = batchMode || parameters.noBuffering();
         try (TerminalOutput output = new TerminalOutput(noBuffering, parameters.rollingWindowSize(), logFile)) {
             try {
                 final ExecutionResult result = new DefaultClient(parameters).execute(output, args);
--- a/client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
+++ b/client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
@@ -33,6 +33,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
 import org.fusesource.jansi.Ansi;
+import org.fusesource.jansi.internal.CLibrary;
 import org.jline.utils.AttributedString;
 import org.jline.utils.AttributedStyle;
 import org.mvndaemon.mvnd.common.BuildProperties;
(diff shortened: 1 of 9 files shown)


gnodet:
I don't think leveraging the somewhat broken (atm) `style.color` option is a good idea, see for example https://issues.apache.org/jira/browse/MNG-6380?filter=12350130.  I've raised a bunch of PR in maven to fix the coloring, but I don't think they'll be integrated in 3.x branch, but rather 4.x.  And I don't really want to wait for maven 4.x for those fixes in `mvnd`.
Also, tying the no-color option to non-buffering does not seem a benefit.  The default buffering mode allows to read the output much more easily (per project logs are grouped together).
What's the problem with the current approach ?

ppalaga:
> I don't think leveraging the somewhat broken (atm) `style.color` option is a good idea, see for example https://issues.apache.org/jira/browse/MNG-6380?filter=12350130. 

With my current proposal, `mvnd -B -Dstyle.color=always package | less -R` outputs colors, as expected. Explicit `-Dstyle.color` is always honored, and `-B` does not override it in any way.

> I've raised a bunch of PR in maven to fix the coloring, but I don't think they'll be integrated in 3.x branch, but rather 4.x. And I don't really want to wait for maven 4.x for those fixes in `mvnd`.
> Also, tying the no-color option to non-buffering does not seem a benefit. 

No, I do not tie no-color to noBuffering. You probably saw this location https://github.com/mvndaemon/mvnd/pull/371/files#diff-2e0b63101a0e22feb7df48ea1a5e83e867c4d076bb72ea4eeccfc5a79dca756bR117 but there, noBuffering is i.a. triggered by `!stdoutIsTerminal` which IMO is the right thing to do. I admit, it might have been a separate commit, because it is not directly related to coloring. I am open to discuss it.

> The default buffering mode allows to read the output much more easily (per project logs are grouped together).

I agree, and do not think I did anything against the benefits the default buffering mode brings.

> What's the problem with the current approach ?

Originally, I had two motivations for these changes (and when implementing another two appeared):

1. I did not like that we add the new `--color` option for something that was solvable by available Maven options. I originally thought we could use `-B` under the hood (and then learned about `-Dstyle.color` which is better). Now that I know that `--color` is proposed for Maven 4, it is actually quite acceptable and if you like, I would not mind adding `--color` back (keeping `-Dstyle.color` as its alias).

2. I found the state after https://github.com/mvndaemon/mvnd/pull/359 a bit ineffective: we instruct the daemon to produce colored messages and then we strip colors in the client https://github.com/mvndaemon/mvnd/pull/359/files#diff-3bfeaf9064d5d02331090c71be3e33960138f55d4fc7cc11fb692a587b30cb91R842 Isn't it more effective to tell the daemon whether we want colored messages upfront using `-Dstyle.color=always` or `-Dstyle.color=never`? You seem to imply that `-Dstyle.color` does not work properly. You have pointed to an example with `-B` and `-Dstyle.color=always` combined, but as I mentioned above, that case works well with the current proposal, IMO. Maybe you know more situations where `-Dstyle.color` does not work as expected?

3. I have found a related bug that existed both before and after https://github.com/mvndaemon/mvnd/pull/359 : Once the daemon was started with `-B` option, all subsequent invocations, albeit without `-B`, were colorless. The current PR solves this problem by always sending an explicit `-Dstyle.color=always` or `-Dstyle.color=never` to the daemon.

4. As noted above, `!stdoutIsTerminal` should trigger `noBuffering`.

3. and 4. are separate problems and I would not mind discussing them separately, after we find an agreement about 1. and 2.

gnodet:
> 1. I did not like that we add the new `--color` option for something that was solvable by available Maven options. I originally thought we could use `-B` under the hood (and then learned about `-Dstyle.color` which is better). Now that I know that `--color` is proposed for Maven 4, it is actually quite acceptable and if you like, I would not mind adding `--color` back (keeping `-Dstyle.color` as its alias).

Ok. I'd rather keep the `--color` option as an alias.

> 2. I found the state after #359 a bit ineffective: we instruct the daemon to produce colored messages and then we strip colors in the client https://github.com/mvndaemon/mvnd/pull/359/files#diff-3bfeaf9064d5d02331090c71be3e33960138f55d4fc7cc11fb692a587b30cb91R842 Isn't it more effective to tell the daemon whether we want colored messages upfront using `-Dstyle.color=always` or `-Dstyle.color=never`? You seem to imply that `-Dstyle.color` does not work properly. You have pointed to an example with `-B` and `-Dstyle.color=always` combined, but as I mentioned above, that case works well with the current proposal, IMO. Maybe you know more situations where `-Dstyle.color` does not work as expected?

We'd have to verify the behavior of MNG-6719

> 3. I have found a related bug that existed both before and after #359 : Once the daemon was started with `-B` option, all subsequent invocations, albeit without `-B`, were colorless. The current PR solves this problem by always sending an explicit `-Dstyle.color=always` or `-Dstyle.color=never` to the daemon.

Yes, that one should be fixed.  My idea was to always have coloring on the server side, but it's true that unless the system property is always set explicitly, it could keep the previous values, which is wrong.

> 4. As noted above, `!stdoutIsTerminal` should trigger `noBuffering`.

I don't think I agree on that one.  I often redirect the output to a log file, but that does not mean I want the project's output to be interleaved.

> 
> Points 3. and 4. are separate problems and I would not mind discussing them separately, after we find an agreement about 1. and 2.

Can we move #4 out.  Also, I think constants would be nice to have for `always`, `never` and `auto` values.

ppalaga:
> > 1. I did not like that we add the new `--color` option for something that was solvable by available Maven options. I originally thought we could use `-B` under the hood (and then learned about `-Dstyle.color` which is better). Now that I know that `--color` is proposed for Maven 4, it is actually quite acceptable and if you like, I would not mind adding `--color` back (keeping `-Dstyle.color` as its alias).
> 
> Ok. I'd rather keep the `--color` option as an alias.

OK, let me add it back.

> > 2. I found the state after #359 a bit ineffective: we instruct the daemon to produce colored messages and then we strip colors in the client https://github.com/mvndaemon/mvnd/pull/359/files#diff-3bfeaf9064d5d02331090c71be3e33960138f55d4fc7cc11fb692a587b30cb91R842 Isn't it more effective to tell the daemon whether we want colored messages upfront using `-Dstyle.color=always` or `-Dstyle.color=never`? You seem to imply that `-Dstyle.color` does not work properly. You have pointed to an example with `-B` and `-Dstyle.color=always` combined, but as I mentioned above, that case works well with the current proposal, IMO. Maybe you know more situations where `-Dstyle.color` does not work as expected?
> 
> We'd have to verify the behavior of MNG-6719

I have verified that `mvn verify | tee mvn.log` when executed on Windows with git (bash) does **not** write the color escape to the file.

> > 3. I have found a related bug that existed both before and after #359 : Once the daemon was started with `-B` option, all subsequent invocations, albeit without `-B`, were colorless. The current PR solves this problem by always sending an explicit `-Dstyle.color=always` or `-Dstyle.color=never` to the daemon.
> 
> Yes, that one should be fixed. My idea was to always have coloring on the server side, but it's true that unless the system property is always set explicitly, it could keep the previous values, which is wrong.

Filed https://github.com/mvndaemon/mvnd/issues/373 so that we have it in the change log.

> > 4 As noted above, `!stdoutIsTerminal` should trigger `noBuffering`.
> 
> I don't think I agree on that one. I often redirect the output to a log file, but that does not mean I want the project's output to be interleaved.

Yes, thanks, that sounds as a valid reason. Let me remove it.

> Also, I think constants would be nice to have for `always`, `never` and `auto` values.

+1

ppalaga:
All done, could you please review?

client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
```diff
@@ -83,24 +84,20 @@ public static void main(String[] argv) throws Exception {
...
+                        Environment.MAVEN_COLOR.removeCommandLineOption(args);
+                    }
+                    return c;
+                }).orElse(Color.auto);
+
+        if (styleColor == Color.auto) {
+            /* stdout is not a terminal e.g. when stdout is redirected to a file */
+            final boolean stdoutIsTerminal = CLibrary.isatty(1) != 0;
+            args.add("-D" + Environment.MAVEN_COLOR.getProperty() + "="
+                    + ((batchMode || logFile != null || !stdoutIsTerminal) ? Color.never.name() : Color.always.name()));
```
gnodet, line 100: That looks a bit complicated.  Can we just remove the option from the command line
```
final Color styleColor = Color.of(Environment.MAVEN_COLOR.removeCommandLineOption(args)).orElse(Color.auto);
// translate auto to either always or never
if (style == Color.auto) {
    /* stdout is not a terminal e.g. when stdout is redirected to a file */
    final boolean stdoutIsTerminal = CLibrary.isatty(1) != 0;
    styleColor = (batchMode || logFile != null || !stdoutIsTerminal) ? Color.never : Color.always;
}
Environment.MAVEN_COLOR.addCommandLineOption(args, styleColor.name());
```

gnodet (changes requested):
There's another thing that comes to my mind.

client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
ppalaga: +1, but we cannot call `Environment.MAVEN_COLOR.addCommandLineOption()` because that would send `--color` to the daemon, where on `-Dstyle.color` is supported

gnodet: Ok, could you add a comment to explain that ? That's not obvious...

ppalaga: +1, done

daemon/src/test/java/org/mvndaemon/mvnd/cache/impl/CacheFactoryTest.java
```diff
@@ -64,6 +66,12 @@ public void assertPutGet(Path tempDir, final Cache<String, CacheRecord> cache, i
         Assertions.assertEquals(record2, cache.get(k2));
         Assertions.assertFalse(record2.invalidated);
 
+        /* Make sure the new content is written a couple of ms later than the original lastModifiedTime */
+        final long deadline = Files.readAttributes(file1, BasicFileAttributes.class).lastModifiedTime().toMillis() + 10;
+        while (System.currentTimeMillis() < deadline) {
+            /* Wait */
+        }
```
ppalaga, line 73: This is my attempt to "fix" the `CacheFactoryTest.timestampCache()` failures on GH actions. They occur only on Linux, and my working hypothesis was that the new content was written to the file so fast after the original content that the timestamp did not change. It is not an issue on Mac, because `BasicFileAttributes.fileKey()` changes when the content is changed. On Linux BasicFileAttributes.fileKey() stays constant for the given path and thus we effectively depend only on the timestamp. I wonder we should come around that? This "fix" is hiding the problem rather than solving it. Do we, under these circumstances, still find `TimestampCacheFactory` good enough for Linux?

gnodet: I'd say that's fine.  I doubt this can really happen as that would mean an artifact is built while being consumed at the same time, which means that either the reactor dependencies are broken, or that the user is building the same project in two different concurrent maven invocation, which is bound to fail anyway, so I'm not too worried about that.

ppalaga: Is this PR good to merge then?

gnodet: @ppalaga The loop should contain a `Thread.sleep(1)` to avoid cpu intensive looping. Otherwise good.

ppalaga:
Added Thread.sleep(1) to CacheFactoryTest

PR #381, opened 2021-03-18
The update of the environment does not work with jdk 16, fixes #380

### LINKED ISSUE #380: The update of the environment does not work with jdk 16
```
➜  ~ jenv shell 16
➜  ~ mvnd -v
mvnd native client 0.4.2-darwin-amd64 (7d488f2c97680fde41508d9442edeabc583ac600)
Terminal: org.jline.terminal.impl.PosixSysTerminal with pty org.jline.terminal.impl.jansi.osx.OsXNativePty
[WARN] Environment mismatch ! Could not set the environment (java.lang.reflect.InaccessibleObjectException: Unable to make field private static final java.util.HashMap java.lang.ProcessEnvironment.theEnvironment accessible: module java.base does not "opens java.lang" to unnamed module @550ee7e5)
[WARN] A few environment mismatches have been detected between the client and the daemon.
[WARN] If the difference matters to you, stop the running daemons using mvnd --stop and
[WARN] start a new daemon from the current environment by issuing any mvnd build command.
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /usr/local/Cellar/mvnd/0.4.2/libexec/mvn
Java version: 16, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/jdk-16.jdk/Contents/Home
Default locale: en_GB, platform encoding: UTF-8
OS name: "mac os x", version: "10.16", arch: "x86_64", family: "mac"
```

gnodet:


gnodet:
> Wouldn't it be easier to use `JDK_JAVA_OPTIONS` env var? https://docs.oracle.com/en/java/javase/11/tools/java.html#GUID-3B1CE181-CD30-4178-9602-230B800D4FAE
> 
> We could perhaps use it unconditionally, bc. it is probably ignored on Java 8.

That would definitely be easier, I'll have a look, thx for the pointer !

ppalaga (approved):
Thanks, this is fine for now.

I wonder how much should we make this configurable for the end users? We could perhaps introduce a new Environment enum item for this, so that end users can customize the default if it does not suit their needs. 

The question is also how smart & complicated should we make the handling of the default? If the end user sets a value without our `--add-opens` options, shouldn't we add it implicitly? If so there could also perhaps be a separate option to disable appending our defaults. WDYT?

gnodet:
> Thanks, this is fine for now.
> 
> I wonder how much should we make this configurable for the end users? We could perhaps introduce a new Environment enum item for this, so that end users can customize the default if it does not suit their needs.
> 
> The question is also how smart & complicated should we make the handling of the default? If the end user sets a value without our `--add-opens` options, shouldn't we add it implicitly? If so there could also perhaps be a separate option to disable appending our defaults. WDYT?

Yes, there's definitely room for improvements and make it more open to the end user.  Right now, the user can always use the `mvnd.jvmArgs` to tailor the JVM options to his needs, so I'm not too worried.

Issue #390, opened 2021-04-02
[Security] Possible RCE


Syquel:
# Attack Vector
mvnd starts a daemon, which listens on a random TCP port ~~in the port range 49152 - 65535 ([IANA Service Name and Transport Protocol Port Number Registry](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml))~~ 1025 - 65535.
The TCP socket is bound to localhost / 127.0.0.1 .

The mvnd daemon utilizes a custom binary protocol on this socket to receive commands like executing Maven builds with a POM file or executing arbitrary Maven plugins. 

An authentication mechanism to verify that the user, which started the mvnd daemon, executes these commands is not available.

# Attack Scenarios
These scenarios are based on the guessability / determinability of the TCP port the mvnd daemon is running on.
~~Due to the limited range of port numbers~~ It is feasible to determine the actual port through port scanning or other means.

## Attack Scenario 1 (Local)
In multi-user environments it is possible for a third-party with a separate user account on the same system to connect to a running mvnd daemon of a victim and to execute arbitrary commands via Maven plugins.

These arbitrary Maven commands a executed with the user permissions of the victim.

### Limitations
- The attacker must have access to the same system as the victim
- The victim must be logged in on the system
- The victim must have executed mvnd recently (within 3 hours by default)

## Attack Scenario 2 (Web Browser)
The attacker is leading on the victim to visit a malicious website (e.g. through social engineering).

Web Browsers do not allow the usage of TCP sockets via JavaScript or WebAssembly, but restrict requests to the HTTP and WebSocket application protocols.
Because the mvnd daemon uses a custom binary protocol, a website is therefore not able to directly interact with the mvnd daemon.

If the victims web browser supports plugins like Java Applet, Flash, Silverlight or similar the website is able to to connect to a running mvnd daemon on the victims system and to execute arbitrary commands via Maven plugins.
These web browsers and web browser plugins usually regard high range ports (49152 - 65535) as secure and therefore do not limit access to them.

These arbitrary Maven commands a executed with the user permissions of the victim on the victims system, which might transcend network boundaries.

### Limitations
- The victim must use a vulnerable web browser, which supports Flash, Silverlight or similar
  - Modern / up-to-date web browsers usually do not support these plugins anymore and Windows disabled Flash on 27.10.2020

### Further Notes
[...]

PR #391, opened 2021-04-02
#390 Restrict usage of mvnd daemons to the current user by utilizing a token check

Syquel:
Adds a token to the persisted daemon registry, which the client is required to transmit directly after initiating the connection with the daemon.

### DIFF
--- a/daemon/src/main/java/org/mvndaemon/mvnd/daemon/Server.java
+++ b/daemon/src/main/java/org/mvndaemon/mvnd/daemon/Server.java
@@ -19,8 +19,11 @@ import java.io.IOException;
 import java.lang.reflect.Field;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.nio.ByteBuffer;
 import java.nio.channels.ServerSocketChannel;
 import java.nio.channels.SocketChannel;
+import java.security.MessageDigest;
+import java.security.SecureRandom;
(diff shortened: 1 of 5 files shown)


daemon/src/main/java/org/mvndaemon/mvnd/daemon/Server.java
```diff
@@ -246,6 +260,23 @@ private void client(SocketChannel socket) {
...
+            do {
+                if (socket.read(tokenBuffer) == -1) {
+                    break;
+                }
+            } while (tokenBuffer.remaining() > 0);
+        } catch (final IOException e) {
+            LOGGER.debug("Discarding EOFException: {}", e.toString(), e);
+        }
+
+        return Arrays.equals(info.getToken(), token);
```
electrum, line 277: This is vulnerable to timing attacks, since the comparison will stop at the first non-equal byte. An attacker can use this to learn the token. Instead, use [MessageDigest.isEqual()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/security/MessageDigest.html#isEqual(byte%5B%5D,byte%5B%5D)) which does a constant time comparison (does not depend on the array contents).

Syquel: Thanks for the hint!
I fixed it.

Issue #384, opened 2021-03-22
Duplicate plugin realm for plugin xxx


lanmaoxinqing:
Hi,
I ran a md5sum check test  between 2000+ real projects built by mvnd-0.4.1 and apache-maven, and glad to say that mvnd build successfully passed most check tests.

5 projects met "Duplicate plugin realm for xxx" exception with mvnd while maven build success, but also succeed in the next recheck.

I find out that `flush` method in `InvalidatingPluginRealmCache` doesn't call `super.flush()`, but I'm not sure whether this is the root case.

Here is the exception stack
```
Caused by: java.lang.IllegalStateException: Duplicate plugin realm for plugin org.apache.maven.plugins:maven-compiler-plugin:3.6.0
        at org.apache.maven.plugin.DefaultPluginRealmCache.put(DefaultPluginRealmCache.java:172)
        at org.mvndaemon.mvnd.cache.invalidating.InvalidatingPluginRealmCache.put(InvalidatingPluginRealmCache.java:106)
        at org.mvndaemon.mvnd.plugin.CliMavenPluginManager.setupPluginRealm(CliMavenPluginManager.java:317)
        at org.apache.maven.plugin.DefaultBuildPluginManager.getPluginRealm(DefaultBuildPluginManager.java:229)
        at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:100)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:208)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
```

ppalaga:
Could you please add some steps to reproduce?

lanmaoxinqing:
I'm sorry it's hard to reproduce. I try to rerun a project 50 times per group in serveral groups and sometimes it appears.

I'm working on the compatiablity of mvnd with maven3.2.5, and find out that I revert some logic in CliMavenPluginManger(setupPluginRealm) back to maven3.2.5, this issue can be ignored before I can guarantee it doesn't matter

lanmaoxinqing:
[...]

Issue #143, opened 2020-10-25
Set maven version like maven-wrapper


herder:
To ensure repeatable builds in a CI environment it would be great to be able specify desired maven version, like the [Maven Wrapper plugin](https://github.com/takari/maven-wrapper). 

Or possibly detect if `mvnw` is present in the project directory and use that?

ppalaga:
It is true that we embed Maven, but we modify it slightly. Pulling a random Maven version during the build and running with it would be rather risky because we test only against a single particular Maven version.

Having our own `mvndw` script would perhaps solve your problem too? I think that kind of solution would be more robust and I'd be open to accept such a contribution.

ppalaga:
Oh, sorry I did not want to close.

herder:
> Providing our own `mvndw` script that would pull a specific version of `mvnd` would perhaps solve your problem too? I think that kind of solution would be more robust and I'd be open to accept such a contribution.

That sounds like a good idea, if it's not possible to use Maven out of the box. 

Adding a `mvnd.version` property that can be passed to the command line or put in `mvnd.properties` would be the easiest way to use this, I guess?

ppalaga:
I thought it could work the very same way as with stock Maven:

* A tool similar to Maven's wrapper plugin (`mvn -N io.takari:maven:0.7.7:wrapper`) would generate `mvndw` and `mvndw.cmd` scripts
* Those scripts would either call `mvnd` already installed somewhere under `~/.m2` (e.g. `/home/ppalaga/.m2/mvnd`) or they would download and install `mvnd` before calling it from there.

soerenhenkel:
Hm. Not sure if I understand that correctly, but this kind of solution would "only" allow to select a specific version of mvnd (like e.g. 0.0.9), wouldn't it? 
I think in the end we would need to be able to select a specific mvn version (like e.g. 3.5.4, instead of 3.6.3 coming with mvnd).

ppalaga:
> we would need to be able to select a specific mvn version (like e.g. 3.5.4, instead of 3.6.3 coming with mvnd).

I am quite sure that supporting older Maven versions is not a priority for me and @gnodet but we are still open to contributions in case anybody else would like to work on that.

herder:
Would it be possible to summarize what modifications you need to do to Maven, @ppalaga ?
I could look it up I guess, but I might miss something :)

ppalaga:
> what modifications you need to do to Maven?

[...]

PR #394, opened 2021-04-07
#393 update Maven version to 3.8.1

### LINKED ISSUE #393: Support Maven 3.8.1
Maven 3.8.1 has been released with some fixes for CVEs:
https://maven.apache.org/docs/3.8.1/release-notes.html

Would it be possible to support this?

robertk3s:
Build locally according to guideline and tested with a local project. OS is Linux.
Is this already enough to verify this is working?

closes #393

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
         <jline.version>3.18.0</jline.version>
         <junit.jupiter.version>5.6.0</junit.jupiter.version>
         <logback.version>1.2.3</logback.version>
-        <maven.version>3.6.3</maven.version>
+        <maven.version>3.8.1</maven.version>
         <slf4j.version>1.7.29</slf4j.version>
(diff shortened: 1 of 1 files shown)


gnodet:
Given 3.8.1 creates some incompatibilities, I'd really like to get the time to work on supporting multiple maven versions as indicated by #384.

robertklaus-bosch:
> 
> Given 3.8.1 creates some incompatibilities, I'd really like to get the time to work on supporting multiple maven versions as indicated by #143 .

Currently, Maven does not release that often. But in general mvnd might require to become more reponsive to release, e.g in case of bugfix releases or Maven 4. So yes, supporting multiple versions might be necessary but I guess this won't be that easy to do, right? And I guess requires more time.

In this regard, what happens with this pull request and by that the Maven 3.8.1 support?

Issue #281, opened 2020-12-21
Could not find artifact org.apache.maven.surefire:surefire-providers:pom:2.22.2


ppalaga:
Happens intermittently on the CI.

```
Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.2:test (default-test) on project multi-module-hello: Unable to generate classpath: org.apache.maven.artifact.resolver.ArtifactResolutionException: Unable to get dependency information for org.apache.maven.surefire:surefire-junit-platform:jar:2.22.2: Failed to process POM for org.apache.maven.surefire:surefire-junit-platform:jar:2.22.2: Non-resolvable parent POM for org.apache.maven.surefire:surefire-junit-platform:2.22.2: Could not find artifact org.apache.maven.surefire:surefire-providers:pom:2.22.2
[ERROR]   org.apache.maven.surefire:surefire-junit-platform:jar:2.22.2
[ERROR] 
[ERROR] from the specified remote repositories:
[ERROR]   central (https://repo.maven.apache.org/maven2, releases=true, snapshots=false)

	at org.apache.maven.repository.legacy.resolver.DefaultLegacyArtifactCollector.recurse(DefaultLegacyArtifactCollector.java:596)
	at org.apache.maven.repository.legacy.resolver.DefaultLegacyArtifactCollector.collect(DefaultLegacyArtifactCollector.java:148)
	at org.apache.maven.artifact.resolver.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:504)
	at org.apache.maven.artifact.resolver.DefaultArtifactResolver.resolveWithExceptions(DefaultArtifactResolver.java:358)
	at org.apache.maven.artifact.resolver.DefaultArtifactResolver.resolveTransitively(DefaultArtifactResolver.java:352)
	at org.apache.maven.artifact.resolver.DefaultArtifactResolver.resolveTransitively(DefaultArtifactResolver.java:324)
	at org.apache.maven.artifact.resolver.DefaultArtifactResolver.resolveTransitively(DefaultArtifactResolver.java:287)
	at org.apache.maven.artifact.resolver.DefaultArtifactResolver.resolveTransitively(DefaultArtifactResolver.java:261)
	at org.apache.maven.plugin.surefire.SurefireDependencyResolver.resolveArtifact(SurefireDependencyResolver.java:130)
[...]

PR #395, opened 2021-04-09
Fix concurrent access to local repository

### LINKED ISSUE #14: Concurrent access to the local repository
Creating this issue to discuss the strategy, because this will hit us sooner or later. It looks like the stock Maven still does not synchronize the access to the local repo. Due to this, two threads or processes can write into the same file which may result in a corrupted data.

https://issues.apache.org/jira/browse/MNG-2802 https://github.com/takari/takari-local-repository

lanmaoxinqing:
remove takari , upgrade maven-resolver version to 1.6.1, try to resolve  #14 #281


lanmaoxinqing:
I ignored the maven-resolver v1.4.1 unzipped from maven-core and copy v1.6.1 into /lib/ext.
It is a pain point that newer maven-resolver versions will no longer automatic upgrade through maven-core.
Any nicer ideas?

gnodet:
> It is a pain point that newer maven-resolver versions will no longer automatic upgrade through maven-core.
> Any nicer ideas?

What do you mean ? I don't understand...

lanmaoxinqing:
> 
> What do you mean ? I don't understand...

Sorry for my poor english... I excludes `maven-resolver*` unzipped from maven-core in `maven-distro.xml`
```xml
    <artifactSet to="/mvn">
        <artifact id="org.apache.maven:apache-maven:tar.gz:bin">
            <unpack useRoot="false"
                    excludes="...lib/maven-resolver-api*,lib/maven-resolver-impl*,lib/maven-resolver-util*,lib/maven-resolver-spi*"/>
        </artifact>
    </artifactSet>
```
and copy v1.6.1 to /mvn/lib/ext
image

so even we upgrade maven which uses higher version of `resolver`, it still will be v1.6.1

gnodet:
I understand now.
I think this PR is obsolete with #393 which includes the resolver 1.6.2 release.
However, I need to check that the takari one is actually not included and that we're actually using the new locking mechanism.

gnodet:
> I understand now.
> I think this PR is obsolete with #393 which includes the resolver 1.6.2 release.
> However, I need to check that the takari one is actually not included and that we're actually using the new locking mechanism.

So upgrading to the new 1.6.2 maven resolver is not sufficient, as the default locking mechanism is just to not lock anything.  There are two implementations: one that only works for in-JVM locking (so between threads), and another one that need a Redis server to run.  So there's currently out-of-the-box option for mvnd.
I'll see if I can write a small dedicated client/server that could be launched by the client if not already started and used by all daemons.

lanmaoxinqing:
> So upgrading to the new 1.6.2 maven resolver is not sufficient, as the default locking mechanism is just to not lock anything. There are two other implementations: one that only works for in-JVM locking (so between threads), and another one that need a Redis server to run. So there's currently out-of-the-box option for mvnd.
> I'll see if I can write a small dedicated client/server that could be launched by the client if not already started and used by all daemons.

You are right, `DefaultSyncContextFactory` in resolver 1.6.2 uses  empty implementation, this PR don't really fix #281 . I'll gonna close it.

PR #399, opened 2021-04-27
Upgrade maven 3.8.1 + prototype for global lock

gnodet:


gnodet:
@michael-o you may want to have a look at this sync context implementation.  It has the benefits of being global to the host without requiring any kind of setup from the user.

michael-o:
Thanks, will take a deep look when we complete 1.7.0 with @cstamas. Hope this will happen until next week.

gnodet:
Draft PR at https://github.com/apache/maven-resolver/pull/105

githubbot-eric:
@gnodet 

Our Jenkins work builders are actually K8s pod based.
I don't think this solution can benefit the container based Jenkins solution.

PR #401, opened 2021-04-27
Provide a native library, fixes #400

### LINKED ISSUE #400: Provide a native library for mvnd specific needs

gnodet:


ppalaga:
@gnodet could you please explain what problem is this solving? There is not info in the linked issue either.

gnodet:
> @gnodet could you please explain what problem is this solving? There is not info in the linked issue either.

Recently, I've added some native methods to jansi specifically for mvnd (`setenv` and `chdir`).  But Jansi has a specific purpose of handling the ansi terminals and it's not supposed to become a location for any kind of native methods.
I've had to add the osx method to get the system memory, so I decided to move them to a more correct location, i.e. inside mvnd.

ppalaga:
Thanks, makes sense!

ppalaga (approved):

Issue #430, opened 2021-06-16
The build time event spy should aggregate values for each mojo


gnodet:
The feature is quite useless when using big aggregator projects as there's no easy way to have an overview of where the time is spent, which is the whole point of this feature.

PR #431, opened 2021-06-16
The build time event spy should aggregate values for each mojo #430

gnodet:
… #430

### DIFF
--- a/daemon/src/main/java/org/mvndaemon/mvnd/timing/BuildTimeEventSpy.java
+++ b/daemon/src/main/java/org/mvndaemon/mvnd/timing/BuildTimeEventSpy.java
@@ -16,6 +16,7 @@
 package org.mvndaemon.mvnd.timing;
 
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


daemon/src/main/java/org/mvndaemon/mvnd/timing/BuildTimeEventSpy.java
```diff
@@ -110,10 +111,16 @@ private void doReport(boolean output) {
         log.accept(DIVIDER);
         log.accept("Build Time Summary:");
         log.accept(DIVIDER);
+        Map<String, Long> mojos = new HashMap<>();
         session.projects().forEach(p -> {
             log.accept(String.format("%s [%.3fs]", p.name(), p.duration() / 1000d));
-            p.mojos().forEach(m -> log.accept(String.format("  %s [%.3fs]", m.name(), m.duration() / 1000d)));
+            p.mojos().forEach(m -> {
+                log.accept(String.format("  %s [%.3fs]", m.name(), m.duration() / 1000d));
+                mojos.merge(m.name(), m.duration(), Long::sum);
```
lanmaoxinqing, line 119: I wonder if we can simply sum the mojo executions durations. Since we parallel build projects, the mojo executions are overlap too. This is not the actual time cost.

gnodet: @lanmaoxinqing I'm not sure to fully understand. Are you saying remove the individual logs ? I.e. remove the first line of the `forEach` or is that something else ?

lanmaoxinqing: Aha still my poor english :-< ...
we calculate the total duration for each mojo by
```
mojos.merge(m.name(), m.duration(), Long::sum);
```
but the actual time we cost is the `endTime` of the last module (e.g. compiler-plugin) - `startTime` of it in first module, this commonly should be less than the total sum cost

gnodet: Still not there...
The current computation is the sum of all runs for a given mojo execution.  If we have 2 modules, the first compilation takes 1s and the other 2s, the merge function will sum these times and will display 3s.  That's what I'm interested in.
It does not display the `(max end) - (min start)` or something like that.

lanmaoxinqing: It's 3s and no problem if we use this computation
> The current computation is the sum of all runs for a given mojo execution. 

but if the second module starts at the same time with first module, the total time cost in real world is 2s. Should we display the sum of all runs, or the time elapsed in builds, that's what i confused.

gnodet: My main usage for this feature is to understand where the time is spent to see if things can be optimized.  So I'm interested in the sum of all runs to see where the CPU time is spent I think.

lanmaoxinqing: I understand now, thank you for your patient reply~

Issue #432, opened 2021-06-17
Bad string size caused by missing bytes in registry.bin


lanmaoxinqing:
Hi,
We occasionally got a `bad string size` exception  in our ci system showing below
```
Exception in thread "main" java.lang.IllegalStateException: Bad string size: 29801
	at org.mvndaemon.mvnd.common.DaemonRegistry.readString(DaemonRegistry.java:274)
	at org.mvndaemon.mvnd.common.DaemonRegistry.doUpdate(DaemonRegistry.java:204)
	at org.mvndaemon.mvnd.common.DaemonRegistry.read(DaemonRegistry.java:159)
	at org.mvndaemon.mvnd.common.DaemonRegistry.getAll(DaemonRegistry.java:101)
	at org.mvndaemon.mvnd.client.DaemonConnector.connect(DaemonConnector.java:97)
	at org.mvndaemon.mvnd.client.DefaultClient.execute(DefaultClient.java:233)
	at org.mvndaemon.mvnd.client.DefaultClient.main(DefaultClient.java:98)
```

In my situation, data keeps correct until reading the missing `short` bytes, which is the length of `reason` field, and then crashes. It may still work if the next `short bytes` is less than 1024 (but the `DaemonStopEvent` data breaks down since then)
```java
    private String readString() {
        int sz = buffer.getShort();
        if (sz == -1) {
            return null;
        }
        if (sz < -1 || sz > 1024) {  // <-- here
            throw new IllegalStateException("Bad string size: " + sz);
        }
        byte[] buf = new byte[sz];
        buffer.get(buf);
        return new String(buf, StandardCharsets.UTF_8);
    }
```
![image](https://user-images.githubusercontent.com/5533593/122327261-2afd8580-cf60-11eb-874f-2cb0b1349b2e.png)
![image](https://user-images.githubusercontent.com/5533593/122327621-bd9e2480-cf60-11eb-8d1c-0167bb0365b8.png)

what's worse, all later builds on the same ci machine will crash due to `readString` exception, until we delete the `registry.bin` file.

It happens in mvnd 0.4.1, but I think we still suffers in 0.5.1.
The problem `registry.bin` is in attach file below (I masked some working path info due to sensitive security policy), be sure using 0.4.1 to debug with it (token field add in 0.5.0)
[registry.bin.zip](https://github.com/mvndaemon/mvnd/files/6667279/registry.bin.zip)

[...]

PR #435, opened 2021-06-17
Temp Workaround to recovery from bad string size, #432

lanmaoxinqing:
Since the registry data is invalid, i simply empty the content to guarantee the next registry will work correctly.


gnodet:
Superseded by #439

Issue #440, opened 2021-06-19
Files still locked after earlier stopped / failed build on Windows


rogierhofboer:
When building openHAB:

mvnd clean install -DskipTests

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-clean-plugin:3.0.0:clean (default-clean) on project org.openhab.core: Failed to clean project: Failed to delete C:\openhab-main\openhab-core\bundles\org.openhab.core\target\org.openhab.core-3.1.0-SNAPSHOT.jar -> [Help 1]

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-clean-plugin:3.0.0:clean (default-clean) on project org.openhab.core.io.http: Failed to clean project: Failed to delete C:\openhab-main\openhab-core\bundles\org.openhab.core.io.http\target\org.openhab.core.io.http-3.1.0-SNAPSHOT.jar -> [Help 1]

The target dir cannot be deleted and is still locked by java

![image](https://user-images.githubusercontent.com/6314129/122640327-4e126b80-d0ff-11eb-95d9-2dd7680e9990.png)

killing the java process fixes the problem, could it be the mvdn does not kill all sub-processes when it is aborted and/or failing to build?

jeffjensen:
Same happens for me on my current product work (no public repo).  It's also reproducible - always the same module.

rogierhofboer:
This seems to be a duplicate of / related to https://github.com/mvndaemon/mvnd/issues/115

gnodet:
I can't build openhab on windows for some reasons (code style issues). Can anyone try with https://github.com/mvndaemon/mvnd/pull/448 ?  It removes the classloader for the `bnd-maven-plugin` from the cache after the build, so it should get rid of the locks hopefully.

rogierhofboer:
Thanks for looking into this. I will try it this evening.

rogierhofboer:
I need a little more time because of other priorities. 
I just succeeded building your i440 branch on Windows with GraalVM / native-image. 
I will try to do some extensive tests tomorrow.

rogierhofboer:
Unfortunately this doesn't solve the problem.

I also have problems getting openHAB to compile on Windows and I am trying to solve them.
Make sure core.autocrlf is set:
`git config --global core.autocrlf true`

And skip all checks and tests (at least for now), by adding the following options (maybe not all of them are needed):
[...]

PR #448, opened 2021-07-09
Provide an eviction pattern to get rid of classloaders for bad behavi…

gnodet:
…ng plugins, #440


kwin:
@gnodet Can you add a bug report at https://github.com/bndtools/bnd and reference the number here?

gnodet:
> @gnodet Can you add a bug report at https://github.com/bndtools/bnd and reference the number here?

I don't think `bnd` exhibits any problem as the actual culprit is the EMF plugin which was spawned using `exec:java`.

PR #451, opened 2021-07-28
Make sure the client env vars are correctly propagated to system properties

gnodet:


ppalaga:
I this good to merge, @gnodet?

We recently upgraded to Maven 3.8.2 and it would be nice to release.

Issue #457, opened 2021-08-21
Support Maven 3.8.2


robertk3s:
Maven 3.8.2 is available with a bunch of bug fixes.
See announcement + release notes here:
http://maven.apache.org/docs/history.html

robertk3s:
Sorry first pr was bogus :/

gnodet:
Maven 3.8.2 has some nasty issues (see https://issues.apache.org/jira/browse/MNG-7216), so I'd rather skip that one and defer the upgrade to 3.8.3.

PR #470, opened 2021-09-07
Revert "#457 Support Maven 3.8.2"

gnodet:
This reverts commit fab202c5f97bcb71a4e5734262eca62395534a73.


ppalaga:
Done because of an performance drop on large projects caused by https://issues.apache.org/jira/browse/MNG-7170 , see also https://issues.apache.org/jira/browse/MNG-7216

Issue #474, opened 2021-09-09
No native library found for os.name=Linux, os.arch=aarch64


lanmaoxinqing:
The `os.arch` system variable may be `aarch64` in some linux arm64 machines like below and `OSInfo.java` in native module doesn't recognize it.
```
$ uname -a
Linux xxx xxx.aarch64 #1 SMP Fri Aug 9 04:09:13 UTC 2019 aarch64 aarch64 aarch64 GNU/Linux
```
I will submit a PR for this later if it's necessary to compatible with this situation.

lanmaoxinqing:
It seems that `org.fusesource.jansi:jansi:2.3.2` which we dependent on also uses `OSInfo.java` and has the same question, So this PR can not cannot actually work until we release a newer jansi version with aarch64 support.
```java
    public static String getArchName() {
        String osArch = System.getProperty("os.arch");
        // For Android
        if (isAndroid()) {
            return "android-arm";
        }

        if (osArch.startsWith("arm")) {    //  <-- linux arch named `aarch64` can not go into this
            osArch = resolveArmArchType();
        } else {
            String lc = osArch.toLowerCase(Locale.US);
            if (archMapping.containsKey(lc))
                return archMapping.get(lc);
        }
        return translateArchNameToFolderName(osArch);
    }
```
https://github.com/fusesource/jansi/blob/master/src/main/java/org/fusesource/jansi/internal/OSInfo.java#L185

PR #475, opened 2021-09-09
Fix native-image mapping for aarch64, #474

lanmaoxinqing:


ppalaga:
There is no way to test this on GH actions yet.

lanmaoxinqing:
> There is no way to test this on GH actions yet.

I'll glad to update GH actions to support aarch64 builds. 

But before that, i find jansi-native also do not support aarch64 due to same problem. 

So this fix cannot actually work until we release a newer jansi version with aarch64 support.

ppalaga:
> > There is no way to test this on GH actions yet.

That was just a side note, not a reason against this, sorry :)

> I'll glad to update GH actions to support aarch64 builds.

I wonder how? AFAIK aarch64 is not supported yet.

lanmaoxinqing:
> That was just a side note, not a reason against this, sorry :)

Oh i just mean that it's my pleasure to update GH actions~ sorry for my poor english :->)

> I wonder how? AFAIK aarch64 is not supported yet.

AFAIK the platform dependent part of maven is native image building and jansi dependency. It will take a few work to compacity with aarch64.

Then just like how we have done in [jansi](https://github.com/fusesource/jansi/blob/master/Makefile), my opinion is using docker image [multiarch/crossbuild](https://hub.docker.com/r/multiarch/crossbuild)  to build and test aarch64 releases in GH actions. 
```
linux-arm64:
	docker run -it --rm -v $$PWD:/workdir -e CROSS_TRIPLE=aarch64-linux-gnu multiarch/crossbuild make clean-native native OS_NAME=Linux OS_ARCH=arm64
```

ppalaga:
> > I wonder how? AFAIK aarch64 is not supported yet.
> 
> AFAIK the platform dependent part of mvnd is native image building and jansi dependency, it only needs a few work to compat with aarch64, and it's my pleasure to figure it out if i'm wrong :)
> 
> Then just like how we have done in [jansi](https://github.com/fusesource/jansi/blob/master/Makefile), my idea is to use docker image [multiarch/crossbuild](https://hub.docker.com/r/multiarch/crossbuild) to build and test aarch64 releases in GH actions.
> 
> ```
> linux-arm64:
> 	docker run -it --rm -v $$PWD:/workdir -e CROSS_TRIPLE=aarch64-linux-gnu multiarch/crossbuild make clean-native native OS_NAME=Linux OS_ARCH=arm64
> ```

Thanks for the explanation (I am really unexperienced with arm builds and testing on Linux)
What you mention covers the building of the aarch64 binary  - that's valuable. 
However, it is still unclear to me whether/how the crossbuild image can help with testing the aarch64 binary on GH Actions?

ppalaga:
Running our tests also on aarch64 would be nice, but it is not a blocker in case there is no way to do it on GH Actions or on some other free CI.

lanmaoxinqing:
> Running our tests also on aarch64 would be nice, but it is not a blocker in case there is no way to do it on GH Actions or on some other free CI.

You are right. The solution above can only test and build binary in docker but not directly in GH actions, which can not block invalid PRs.

Specify a arm64 machine to run GH actions will be the best solution, but it seems that only self-hosted machines is supported so far. 

* [adding-self-hosted-runners](https://docs.github.com/en/actions/hosting-your-own-runners/adding-self-hosted-runners)
* [using-github-actions-to-test-your-go-code-on-arm64](https://medium.com/@nabil.servais/using-github-actions-to-test-your-go-code-on-arm64-ce581e646cb)

electrum:
You can use https://github.com/docker/setup-qemu-action to allow running `arm64` Docker images on GH actions:
```yaml
- uses: docker/setup-qemu-action@v1
  with:
    platforms: arm64

lanmaoxinqing:
> You can use https://github.com/docker/setup-qemu-action to allow running `arm64` Docker images on GH actions:

That’s great! I will try it later when i have finish my current works, thanks for your help~

PR #480, opened 2021-09-23
Provide a local (semaphore based) sync context and use it as the default

gnodet:


daemon/src/main/java/org/mvndaemon/mvnd/sync/MvndSyncContextFactory.java
```diff
@@ -80,4 +90,69 @@ public void close() {
...
+                String current = null;
+                try {
+                    for (String key : keys) {
+                        current = key;
+                        getLock(key).lock();
+                        locked.add(key);
+                    }
+                } catch (Exception e) {
+                    close();
+                    throw new IllegalStateException("Could not acquire lock for '" + current + "'" );
```
ppalaga, line 118: Maybe we should not swallow the exception?

```suggestion
                    throw new IllegalStateException("Could not acquire lock for '" + current + "'", e);
```

daemon/src/main/java/org/mvndaemon/mvnd/sync/MvndSyncContextFactory.java
```diff
@@ -80,4 +90,69 @@ public void close() {
...
+        }
+
+        private class LocalSyncContext implements SyncContext {
+            private final Deque<String> locked = new ArrayDeque<>();
+
+            @Override
+            public void acquire(Collection<? extends Artifact> artifacts, Collection<? extends Metadata> metadatas) {
+                Collection<String> keys = new TreeSet<>();
+                stream(artifacts).map(this::getKey).forEach(keys::add);
+                stream(metadatas).map(this::getKey).forEach(keys::add);
```
ppalaga, line 108: I wonder whether we really need the keys collection and esp. TreeSet? Is sorting and uniqueness important? I guess for the ordering, it only matters that the processing order is reproducible, so we can take over the iteration ordering of the passed in collections. For uniqueness, I'd expect the callers to take care, but even if they don't, ReentrantLock can do well with multiple subsequent lock() and unlock() calls from the same thread. So I think we could avoid allocating the key Set and have two loops for artifacts and metadata. WDYT?

PR #481, opened 2021-09-23
Download required native JNI headers

gnodet:


ppalaga:
Hm... is it really necessary to have the library binaries under native/src/main/resources/org/mvndaemon/mvnd/nativ in git? Are they not produced by the build?

ppalaga:
> > Hm... is it really necessary to have the library binaries under native/src/main/resources/org/mvndaemon/mvnd/nativ in git? Are they not produced by the build?
> 
> No, they're not actually part of the build.

Do we have documented somewhere where they come from or how to build them?

PR #502, opened 2021-10-19
Support mac-arm64 platform

gnodet:


gnodet:
There's no separate release.  GraalVM does not support mac-arm64 so there's no way we can generate a specific distribution for this platform.  However, the darwin-amd64 distribution should work on M1 mac, when running with the rosetta translator.

breun:
What does ‘support mac-arm64 platform’ then mean exactly?

PR #507, opened 2021-10-20
Update mvnd to include Maven Resolver 1.7

cstamas:
As mvnd itself is Java8 lang level, it is completely
okay to up resolver to 1.7 (that is java8 as well)
and get it's improvements.

The only reason why Maven 3.8.x CANNOT use resolver
1.7 is that it is still Java7 level.


cstamas:
No, that's misunderstanding. There are several NamedLock providers and only SOME of them needs extra infra (external process):
- `rwlock-local` (default), uses JVM `ReentrantReadWriteLock` per lock name, usable for MT builds.
- `semaphore-local`, uses JVM `Semaphore` per lock name, usable for MT builds.
- `noop`, implement no-op locking (no locking). For experimenting only. Has same functionality as old "nolock"
  SyncContextFactory implementation.

and the pointed PR adds `file-lock` provider using NIO FileLock. **None of these above needs anything external (infra/process)**.

There are TWO providers, for Redis and Hazelcast, and those two DO need external Redis or Hazelcast.

gnodet:
> No, that's misunderstanding. There are several NamedLock providers and only SOME of them needs extra infra (external process):
> 
> * `rwlock-local` (default), uses JVM `ReentrantReadWriteLock` per lock name, usable for MT builds.
> * `semaphore-local`, uses JVM `Semaphore` per lock name, usable for MT builds.
> * `noop`, implement no-op locking (no locking). For experimenting only. Has same functionality as old "nolock"
>   SyncContextFactory implementation.
> 
> and the pointed PR adds `file-lock` provider using NIO FileLock usable for MP. **None of these above needs anything external (infra/process)**.
> 
> There are TWO MP providers, for Redis and Hazelcast, and those two DO need external Redis or Hazelcast.

Yes, what I meant is that in the context of `mvnd`, those two are the only ones that can provide locking between the daemons.   The file lock seems really promising though.

ppalaga:
@gnodet is this good to merge? 
@cstamas could you please squash your commits and possibly rebase on top of recent `main`?

cstamas:
@ppalaga done

PR #508, opened 2021-10-21
Mvnd with file locking

cstamas:
With resolver 1.7.3 new FS locking is available. This PR makes mvnd use it to synchronize (no external bits needed).

Locally, it performs even better than IPC locking (built maven master, not bench-marking, just looking at wall time of Mvn output log).

This PR is here just for experimenting, as:
- it drops IPC sync module from mvnd and syncContextFactory related bits
- introduces own selector with defaults other than mvn


cstamas:
Note: https://github.com/apache/maven-resolver/pull/131 is being prepped for merge, so "hack" (to make FS based locks default) is removed from it, hence the PR description to build mvnd will NOT work anymore (you need to undo in mvnd code these changes to set system properties OR to completely replace selector implementation https://github.com/apache/maven-resolver/pull/131/commits/ef890d75ee8f99fdcaada1f071247bec7d17769d)

cstamas:
https://github.com/apache/maven-resolver/pull/131 is merged.
It depends now on https://github.com/apache/maven-resolver/pull/138

daemon/src/main/java/org/mvndaemon/mvnd/syncontext/DaemonNamedLockFactorySelector.java
```diff
@@ -0,0 +1,61 @@
...
+import org.eclipse.sisu.Priority;
+
+/**
+ * Mvnd selector implementation: it differs from
+ * {@link org.eclipse.aether.internal.impl.synccontext.named.SimpleNamedLockFactorySelector} only by default values.
+ */
+@Singleton
+@Named
+@Priority(10)
+public final class DaemonNamedLockFactorySelector
```
michael-o, line 54: So this is quite nice, the priority will replace the one bundled with Resolver!

daemon/src/main/java/org/mvndaemon/mvnd/syncontext/DaemonNamedLockFactorySelector.java
```diff
@@ -0,0 +1,61 @@
...
+        extends NamedLockFactorySelectorSupport {
+    @Inject
+    public DaemonNamedLockFactorySelector(final Map<String, NamedLockFactory> factories,
+            final Map<String, NameMapper> nameMappers) {
+        super(factories, FileLockNamedLockFactory.NAME, nameMappers, FileGAVNameMapper.NAME);
```
michael-o, line 59: This needs to be documented that mvnd requires a FS underneath with FS advisory lock support. Likely NFS won't do.

michael-o (approved):

PR #524, opened 2021-12-08
Upgrade to maven 3.8.4

gnodet:


findepi:
Hi, thanks for this change! Looking forward to seeing it released, as I wanted to use mvnd to build a project that requires maven 3.8.4 already (which happened to be https://github.com/apache/spark today).

gnodet:
> Hi, thanks for this change! Looking forward to seeing it released, as I wanted to use mvnd to build a project that requires maven 3.8.4 already (which happened to be https://github.com/apache/spark today).

It seems to me that the requirement for spark should not have been bumped to 3.8.4, there's no reason for that.

Issue #527, opened 2021-12-10
Build hang and CPU skyrocket with sisu-index:indexMojo


cstamas:
Unsure (yet :smile: ) who to blame, so just linking the issue here as well:
https://github.com/eclipse/sisu.inject/issues/38

cstamas:
Would be thankful for some pointers, but this seems mvnd related issue.

Using: Linux, mvn 3.8.4, mvnd 0.7.1, java11-temurin

Facts in short: on a closed source project, one module (that uses `sisu-index:index` mojo) build takes 9 seconds with vanilla mvn, while it takes 2:20 with mvnd, and mvnd build is stuck on `sisu-index:index` mojo for 2 minutes.

The setup of project is somewhat similar to test project https://github.com/cstamas/sisu-index-test but test project is trivial and cannot reproduce the issue for now with it (or the execution times are so small that issue is invisible): several plain JAR modules w/ sisu components and then a war-like module (not using war packaging but m-assembly-p) that re-indexes all sisu components in included JARs).

By using (locally patched) sisu-maven-plugin verified that in both cases (executing with mvn and with mvnd) they "do same thing" -- the indexPath and classPath are 100% same (made these two methods dump their list of URLs https://github.com/eclipse/sisu.mojos/blob/master/src/main/java/org/eclipse/sisu/mojos/IndexMojo.java#L176 and https://github.com/eclipse/sisu.mojos/blob/master/src/main/java/org/eclipse/sisu/mojos/IndexMojo.java#L188).

Still, for some reason while running in mvnd, this takes 2:20 total time, while in mvn it finishes in 9 seconds.

cstamas:
Found the culprit:
https://github.com/mvndaemon/mvnd/blob/master/common/src/main/java/org/mvndaemon/mvnd/common/MavenDaemon.java#L31

Adding same change to mvn/index mojo makes vanilla mvn run same goal for 2:20 as well...

PR #529, opened 2021-12-11
[Proposal] Dispose CL and caches per daemon run

cstamas:
Coming from here https://github.com/mvndaemon/mvnd/issues/527

Disabling URL connection caches globally is like
using shotgun against a fly.

Instead, re-create a "disposable" classloader per
Maven Daemon CLI run, and clear caches once
done.

This still keeps JVM "warm", but introduces some
penalty to reconstruct maven from scratch, but
will flush things like once loaded extensions,
keeping work more correct.

### DIFF
--- a/common/src/main/java/org/mvndaemon/mvnd/common/ClassLoaderHelper.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/ClassLoaderHelper.java
@@ -25,18 +25,18 @@ import java.util.stream.Stream;
 
 import static java.util.Objects.requireNonNull;
 
-public class ClassLoaderHelper {
+public final class ClassLoaderHelper {
 
     private ClassLoaderHelper() {
         // no instances
     }
 
     /**
-     * Creates mvndaemon loader based on mvnd home (and known layout).
+     * Creates mvndaemon class loader based on {@link Environment#MVND_HOME} and known layout.
      *
      * @param  filter non-null filter applied to findClass method.
-     * @param  parent nullable parent classloader.
-     * @return        a {@link ClassLoader} that uses passes in filter and parent.
+     * @param  parent nullable parent class loader.
+     * @return        a built {@link ClassLoader} that uses passed in filter and optionally parent.
      */
     public static ClassLoader createLoader(final Predicate<String> filter, final ClassLoader parent) {
 
--- a/common/src/main/java/org/mvndaemon/mvnd/common/ClassLoaderHelper.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/ClassLoaderHelper.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.mvndaemon.mvnd.common;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+import static java.util.Objects.requireNonNull;
+
+public class ClassLoaderHelper {
+
+    private ClassLoaderHelper() {
+        // no instances
+    }
+
+    /**
+     * Creates mvndaemon loader based on mvnd home (and known layout).
+     *
+     * @param  filter non-null filter applied to findClass method.
+     * @param  parent nullable parent classloader.
+     * @return        a {@link ClassLoader} that uses passes in filter and parent.
+     */
+    public static ClassLoader createLoader(final Predicate<String> filter, final ClassLoader parent) {
+
+        requireNonNull(filter);
+        final Path mvndHome = Environment.MVND_HOME.asPath();
+        URL[] classpath = Stream.concat(
+                /* jars */
+                Stream.of("mvn/lib/ext", "mvn/lib", "mvn/boot")
+                        .map(mvndHome::resolve)
+                        .flatMap((Path p) -> {
+                            try {
+                                return Files.list(p);
+                            } catch (java.io.IOException e) {
+                                throw new RuntimeException("Could not list " + p, e);
+                            }
+                        })
+                        .filter(p -> {
+                            final String fileName = p.getFileName().toString();
+                            return fileName.endsWith(".jar") && !fileName.startsWith("mvnd-client-");
+                        })
+                        .filter(Files::isRegularFile),
+                /* resources */
+                Stream.of(mvndHome.resolve("mvn/conf"), mvndHome.resolve("mvn/conf/logging")))
+
+                .map(Path::normalize)
+                .map(Path::toUri)
+                .map(uri -> {
+                    try {
+                        return uri.toURL();
+                    } catch (MalformedURLException e) {
+                        throw new RuntimeException(e);
+                    }
+                })
+                .toArray(URL[]::new);
+        return new URLClassLoader(classpath, parent) {
+
+            private final ClassLoader fallback = parent != null ? parent : ClassLoaderHelper.class.getClassLoader();
+
+            @Override
+            protected Class<?> findClass(String name) throws ClassNotFoundException {
+                try {
+                    if (filter.test(name)) {
+                        return super.findClass(name);
+                    }
+                } catch (ClassNotFoundException e) {
+                    return fallback.loadClass(name);
+                }
+                throw new ClassNotFoundException(name);
+            }
+
+            @Override
+            public URL getResource(String name) {
+                URL url = null;
+                if (filter.test(name)) {
+                    url = super.getResource(name);
+                }
+                if (url == null) {
+                    url = fallback.getResource(name);
+                }
+                return url;
+            }
+        };
+    }
+}
--- a/common/src/main/java/org/mvndaemon/mvnd/common/MavenDaemon.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/MavenDaemon.java
@@ -15,66 +15,10 @@
  */
 package org.mvndaemon.mvnd.common;
 
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.net.URLClassLoader;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.stream.Stream;
-
 public class MavenDaemon {
 
     public static void main(String[] args) throws Exception {
-
-        final Path mvndHome = Environment.MVND_HOME.asPath();
-        URL[] classpath = Stream.concat(
-                /* jars */
-                Stream.of("mvn/lib/ext", "mvn/lib", "mvn/boot")
-                        .map(mvndHome::resolve)
-                        .flatMap((Path p) -> {
-                            try {
-                                return Files.list(p);
-                            } catch (java.io.IOException e) {
-                                throw new RuntimeException("Could not list " + p, e);
-                            }
-                        })
-                        .filter(p -> {
-                            final String fileName = p.getFileName().toString();
-                            return fileName.endsWith(".jar") && !fileName.startsWith("mvnd-client-");
-                        })
-                        .filter(Files::isRegularFile),
-                /* resources */
-                Stream.of(mvndHome.resolve("mvn/conf"), mvndHome.resolve("mvn/conf/logging")))
-
-                .map(Path::normalize)
-                .map(Path::toUri)
-                .map(uri -> {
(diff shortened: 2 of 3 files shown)


cstamas:
This makes originally linked issue behave same as with vanilla mvn, but also makes extensions "flushed". 

Extension problem: use released mvnd 0.7.1 and just build this project https://github.com/cstamas/sisu-index-test
Once built, whatever ELSE you build with mvnd (using same daemon instance), you will have takari/maven-timeline applied to it, even if built project does NOT defines it, as that project above made it loaded up (and remains "resident")

cstamas:
@gnodet pls look at this, as this change works for me great. Unsure how to properly test removed bit, or create some test for it:
```
        // Disable URL caching so that  the JVM does not try to cache resources
        // loaded from jars that are built by a previous run
        new File("txt").toURI().toURL().openConnection().setDefaultUseCaches(false);
```

Also, maybe instead of classloader trickery, more fine grained separation of modules would be needed?

ppalaga:
I'd vote for having "Dispose CL and caches per daemon run" as a configurable feature that would be off by default. I think we should generally push towards fixing plugins to work well over multiple invocations from a single classloader.

As a related but still different task, we should aim at using a better classloade impl than URLClassloader so that the classloader disposal works correctly.

cstamas:
@ppalaga Agree with your assessment, but here are two issues (sorry to mix them up):
issueNo1 = mvnd doing `new File("txt").toURI().toURL().openConnection().setDefaultUseCaches(false);` that made me arrive here (see https://github.com/mvndaemon/mvnd/issues/527 and https://github.com/eclipse/sisu.inject/issues/38)
issueNo2 = "resident" extensions (and who knows what more): while chasing issueNo1 I used some extensions (see https://github.com/cstamas/sisu-index-test/blob/master/pom.xml#L128-L134 ) that from one moment got applied to ALL mvnd built stuff, not only the test project.

Hence this PR as "draft" as am myself not really convinced is this right thing to do (or, how to test this)

gnodet:
I've been looking at this yesterday.

For the resident extension problem, I'm still trying to figure out where in sisu the thread classloader is used.  My understanding is, for the maven timeline extension, the extension provides a `AbstractMavenLifecycleParticipant` bean.  Those are retrieved using the `DefaultMaven#getLifecycleParticipants` method (https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/DefaultMaven.java#L466-L511).  A first scan is done using the container's classloader and then one scan with each the classloader of each extension.  However, the second time this method is called, the first call to `container.lookupList` returns all beans that were found in the previous run.  I wonder if there's a problem here...  It seems to me the calls should be idempotent.

cstamas:
> ...I think we should generally push towards fixing plugins to work well over multiple invocations from a single classloader.

This is all due Plexus container: Cli will keep container up, and maven extensions are actually components loaded up by Plexus as directed by maven. So, as Maven does not do any container "cleanup" per se (as it's JVM lifecycle matches CLI invocation), we may want to look into removing components after build end.... or just rebuild Plexus from scratch, as this PR does.

cstamas:
> For the resident extension problem, I'm still trying to figure out where in sisu the thread classloader is used. My understanding is, for the maven timeline extension, the extension provides a `AbstractMavenLifecycleParticipant` bean. Those are retrieved using the `DefaultMaven#getLifecycleParticipants` method (https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/DefaultMaven.java#L466-L511). A first scan is done using the container's classloader and then one scan with each the classloader of each extension. However, the second time this method is called, the first call to `container.lookupList` returns all beans that were found in the previous run. I wonder if there's a problem here... It seems to me the calls should be idempotent.

No, they are components loaded up by plexus, and they remain in container (until container shut down or each component disposed). This is due daemon holding on container AFTER build is done, as maven per se does not remove anything AFAIK at the end of the build.

We could change CLI to accept "parent" container (+ realm) and make CLI actually create child realm and not "root" contaier? So unlike this PR always reconstruct container from scratch, just create child, while most of Maven is in "parent".... still, unsure would it work, but I see no obstacle from sisu (and plexus-shim) side, unless I missing something.

cstamas:
Am convinced that destroying Plexus once daemon build is done is correct thing to do (as otherwise "fixing things" would not mean only fixing maven, but ANY extension out there that may do wild things as they have access to container). Question IMHO is only do we want to reconstruct from scratch (as this PR) or be smarter and do something about it?

ppalaga:
I have recently seen (see around here https://github.com/quarkusio/quarkus/pull/21719#discussion_r762771096 ) slight differences in how EventSpies and AbstractMavenLifecycleParticipants are registered in mvn vs mvnd. On mvnd EventSpies are always registered, even without `<extensions>true</extensions>` - thats not the case with stock mvn. On stock mvn, maven-core is not available in a plugin's classloader unless the plugin is declared with `<extensions>true</extensions>`. Maybe this plays some role also here?

gnodet:
I don't see how reconstructing the whole container would allow t keep some cache as all the caches are currently singletons and will be destroyed.  Also, even the classloaders will be destroyed I think. If that's the case, the JVM is not *hot* anymore, as all the classes will have to be reloaded, which removes a lot of the benefits of `mvnd`.  I suppose we can measure things.

For the resident extension, I think there's a simple fix, as the visibility is currently wrong, and that's because the TCCL is not set correctly I think.  For the first issue, we may already have a workaround with https://github.com/mvndaemon/mvnd/issues/440.  That's not perfect, but I think it's the same kind of issues with mojos not supporting concurrent builds and maven displaying a warning when the mojo is not annotated with `@threadsafe`.  The difference is that we *can* do something about it in `mvnd` by getting rid of the classloader.

Btw, I haven't looked deeper, but how does the url cache affects sisu index ?

cstamas:
Well, if you do not unload (by disposing container or evicting extension) but keep loading extension with newer and newer class realm/class loader, you will introduced a leak to daemon, no? As each build will load (even the same) extension, but nothing will evict it from JVM... So this may fix the issue (to not execute extension when not needed) but will keep it in memory, no?  But re caches, yes, that's true.

Re url cache: I am really unsure still WHY it affects sisu, but I do know HOW it affects it, what is the outcome: with `mvn` the `maven-sisu-index:index` goal takes 9 seconds, while with `mvnd` same goal (even if single module executed!) takes over 2:20. With this PR (having that line removed), `mvnd` does it in 9 seconds just like `mvn`.

gnodet:
> Well, if you do not unload (by disposing container or evicting extension) but keep loading extension with newer and newer class realm/class loader, you will introduced a leak to daemon, no? As each build will load (even the same) extension, but nothing will evict it from JVM... So this may fix the issue (to not execute extension when not needed) but will keep it in memory, no? But re caches, yes, that's true.

Yes, though the cache can easily be enhanced to use an LRU mechanism or whatever.  Also daemons that runs out of memory do expire and are not reused.

Fwiw, I've fixed the resident extension problem by setting the TCCL correctly with https://github.com/mvndaemon/mvnd/pull/531.  We definitely need an integration test for that.  We definitely need an IT for that though.
 
> Re url cache: I am really unsure still WHY it affects sisu, but I do know HOW it affects it, what is the outcome: with `mvn` the `maven-sisu-index:index` goal takes 9 seconds, while with `mvnd` same goal (even if single module executed!) takes over 2:20 (so 15x slower that is too much IMO). With this PR (having that line removed), `mvnd` does it in 9 seconds just like `mvn`.

cstamas:
Ok, seems I was wrong: as realms are (stable) named/keyed, there is no realm leak (built locally modded mvnd to log out realms on begin and end of org.apache.maven.cli.DaemonMavenCli#doMain) and built https://github.com/cstamas/sisu-index-test 3 times subsequently. 

```
1st run

BEGIN Class realms: plexus.core,maven.ext,maven.api

END Class realms: plexus.core,maven.ext,maven.api,plugin>org.apache.maven.plugins:maven-clean-plugin:3.1.0,plugin>org.eclipse.sisu:sisu-maven-plugin:0.3.5,plugin>org.apache.maven.plugins:maven-install-plugin:2.5.2,plugin>org.apache.maven.plugins:maven-resources-plugin:3.0.2,plugin>org.apache.maven.plugins:maven-compiler-plugin:3.8.0,plugin>org.apache.maven.plugins:maven-surefire-plugin:2.22.1,plugin>org.apache.maven.plugins:maven-jar-plugin:3.0.2,plugin>org.apache.maven.plugins:maven-assembly-plugin:3.3.0,extension>org.apache.felix:maven-bundle-plugin:2.5.0,project>com.google.guava:guava:27.1-jre,extension>org.apache.felix:maven-bundle-plugin:2.4.0,extension>org.sonatype.plugins:nexus-staging-maven-plugin:1.6.3,project>com.google.code.findbugs:jsr305:3.0.2,extension>org.eclipse.tycho:tycho-maven-plugin:0.22.0,project>org.eclipse.sisu:org.eclipse.sisu.inject:0.3.5


2nd run

BEGIN Class realms: plexus.core,maven.ext,maven.api,plugin>org.apache.maven.plugins:maven-clean-plugin:3.1.0,plugin>org.eclipse.sisu:sisu-maven-plugin:0.3.5,plugin>org.apache.maven.plugins:maven-install-plugin:2.5.2,plugin>org.apache.maven.plugins:maven-resources-plugin:3.0.2,plugin>org.apache.maven.plugins:maven-compiler-plugin:3.8.0,plugin>org.apache.maven.plugins:maven-surefire-plugin:2.22.1,plugin>org.apache.maven.plugins:maven-jar-plugin:3.0.2,plugin>org.apache.maven.plugins:maven-assembly-plugin:3.3.0,extension>org.apache.felix:maven-bundle-plugin:2.5.0,project>com.google.guava:guava:27.1-jre,extension>org.apache.felix:maven-bundle-plugin:2.4.0,extension>org.sonatype.plugins:nexus-staging-maven-plugin:1.6.3,project>com.google.code.findbugs:jsr305:3.0.2,extension>org.eclipse.tycho:tycho-maven-plugin:0.22.0,project>org.eclipse.sisu:org.eclipse.sisu.inject:0.3.5

END Class realms: plexus.core,maven.ext,maven.api,plugin>org.apache.maven.plugins:maven-clean-plugin:3.1.0,plugin>org.eclipse.sisu:sisu-maven-plugin:0.3.5,plugin>org.apache.maven.plugins:maven-install-plugin:2.5.2,plugin>org.apache.maven.plugins:maven-resources-plugin:3.0.2,plugin>org.apache.maven.plugins:maven-compiler-plugin:3.8.0,plugin>org.apache.maven.plugins:maven-surefire-plugin:2.22.1,plugin>org.apache.maven.plugins:maven-jar-plugin:3.0.2,plugin>org.apache.maven.plugins:maven-assembly-plugin:3.3.0,extension>org.apache.felix:maven-bundle-plugin:2.5.0,project>com.google.guava:guava:27.1-jre,extension>org.apache.felix:maven-bundle-plugin:2.4.0,extension>org.sonatype.plugins:nexus-staging-maven-plugin:1.6.3,project>com.google.code.findbugs:jsr305:3.0.2,extension>org.eclipse.tycho:tycho-maven-plugin:0.22.0,project>org.eclipse.sisu:org.eclipse.sisu.inject:0.3.5


3rd run

BEGIN Class realms: plexus.core,maven.ext,maven.api,plugin>org.apache.maven.plugins:maven-clean-plugin:3.1.0,plugin>org.eclipse.sisu:sisu-maven-plugin:0.3.5,plugin>org.apache.maven.plugins:maven-install-plugin:2.5.2,plugin>org.apache.maven.plugins:maven-resources-plugin:3.0.2,plugin>org.apache.maven.plugins:maven-compiler-plugin:3.8.0,plugin>org.apache.maven.plugins:maven-surefire-plugin:2.22.1,plugin>org.apache.maven.plugins:maven-jar-plugin:3.0.2,plugin>org.apache.maven.plugins:maven-assembly-plugin:3.3.0,extension>org.apache.felix:maven-bundle-plugin:2.5.0,project>com.google.guava:guava:27.1-jre,extension>org.apache.felix:maven-bundle-plugin:2.4.0,extension>org.sonatype.plugins:nexus-staging-maven-plugin:1.6.3,project>com.google.code.findbugs:jsr305:3.0.2,extension>org.eclipse.tycho:tycho-maven-plugin:0.22.0,project>org.eclipse.sisu:org.eclipse.sisu.inject:0.3.5

END Class realms: plexus.core,maven.ext,maven.api,plugin>org.apache.maven.plugins:maven-clean-plugin:3.1.0,plugin>org.eclipse.sisu:sisu-maven-plugin:0.3.5,plugin>org.apache.maven.plugins:maven-install-plugin:2.5.2,plugin>org.apache.maven.plugins:maven-resources-plugin:3.0.2,plugin>org.apache.maven.plugins:maven-compiler-plugin:3.8.0,plugin>org.apache.maven.plugins:maven-surefire-plugin:2.22.1,plugin>org.apache.maven.plugins:maven-jar-plugin:3.0.2,plugin>org.apache.maven.plugins:maven-assembly-plugin:3.3.0,extension>org.apache.felix:maven-bundle-plugin:2.5.0,project>com.google.guava:guava:27.1-jre,extension>org.apache.felix:maven-bundle-plugin:2.4.0,extension>org.sonatype.plugins:nexus-staging-maven-plugin:1.6.3,project>com.google.code.findbugs:jsr305:3.0.2,extension>org.eclipse.tycho:tycho-maven-plugin:0.22.0,project>org.eclipse.sisu:org.eclipse.sisu.inject:0.3.5

```
With https://github.com/mvndaemon/mvnd/pull/531 then only https://github.com/mvndaemon/mvnd/issues/527 remains, so am closing this out.

PR #534, opened 2021-12-15
Improve the doc text of mvnd.pluginRealmEvictPattern option, fixes #533

### LINKED ISSUE #533: Improve the doc text of `mvnd.pluginRealmEvictPattern` option
From the current text https://github.com/mvndaemon/mvnd/blob/4bb5c5d5b533938d7710eeb08ac1aff5800e29d5/common/src/main/java/org/mvndaemon/mvnd/common/Environment.java#L253-L256 it is not clear how to use the option. An example would help. The option says it is a regexp, but based on https://github.com/gnodet/mvnd/commit/9ed8647014c311e5711bb8c40afa7f6abc4ea5f4#diff-e65744bdec30df4a8c9692d35f97c28fef1e81af93a54a3d0dbf5f0a8e8abd88R246 it seems to be that glob is the default.

gnodet:


common/src/main/java/org/mvndaemon/mvnd/common/Environment.java
```diff
@@ -252,11 +252,12 @@
...
-     *   <li>a regex pattern starting with <code>regex:</code></li>
-     *   <li>a maven expression, either <code>mvn:[groupId]:[artifactId]:[version]</code>,
-     *      <code>mvn:[groupId]:[artifactId]</code> or <code>mvn:[artifactId]</code></li>
+     * The overall pattern is a comma separated list of:
+     * <ul>
+     * <li>a glob pattern starting with <code>glob:</code> (the default syntax if no scheme is specified)</li>
+     * <li>a regex pattern starting with <code>regex:</code></li>
+     * <li>a maven expression, either <code>mvn:[groupId]:[artifactId]:[version]</code>,
+     * <code>mvn:[groupId]:[artifactId]</code> or <code>mvn:[artifactId]</code></li>
      * </ul>
```
ppalaga, line 261: Does the HTML and JavaDoc `{@literal /}` formatting look good in `mvnd --help`?

gnodet: It should now:
```
-Dmvnd.pluginRealmEvictPattern=<string> Pattern that will force eviction of the plugin realms if one of its dependencies matches. The overall pattern is a
                                        comma separated list of either: a glob pattern starting with 'glob:' (the default syntax if no scheme is
                                        specified), a regex pattern starting with 'regex:', a maven expression, either
                                        'mvn:[groupId]:[artifactId]:[version]', 'mvn:[groupId]:[artifactId]' or 'mvn:[artifactId]'. This pattern will be
                                        evaluated against the full path of the dependencies, so it is usually desirable to start with 'glob:**/' to
                                        support any location of the local repository.
```

ppalaga: Thanks, looks good!

I wonder whether a path against the pattern is matched can also be outside of the local maven repo? - Some sort of `$PWD/target/my.jar` is not possible?

Is this only matched against jar paths? Is there no chance that class directories are a part some classloader?

gnodet: For plugins, artifacts are always resolved and thus come from the local repository, so I don't think they can include things like `$PWD/target/my.jar` or `target/classes`.

PR #538, opened 2021-12-16
Fastclean extension

gnodet:
Fast clean extension.

It needs to be registered as an extension, so either in `.mvn/extensions` with
```
<extensions>
    <extension>
        <groupId>org.mvndaemon.mvnd</groupId>
        <artifactId>mvnd-fastclean-maven-plugin</artifactId>
        <version>0.7.2-SNAPSHOT</version>
    </extension>
</extensions>
```

or inside the pom with:
```
<build>
    <extensions>
      <extension>
        <groupId>org.mvndaemon.mvnd</groupId>
        <artifactId>mvnd-fastclean-maven-plugin</artifactId>
        <version>0.7.2-SNAPSHOT</version>
      </extension>
    </extensions>
</build>
```

@atlib77 let me know how this works for you.

### DIFF
--- a/fastclean/pom.xml
+++ b/fastclean/pom.xml
@@ -1,0 +1,113 @@
+<!--
+
+    Copyright 2021 the original author or authors.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+         http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.mvndaemon.mvnd</groupId>
+        <artifactId>mvnd</artifactId>
+        <version>0.7.2-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>mvnd-fastclean-maven-plugin</artifactId>
+
+    <packaging>maven-plugin</packaging>
+    <name>Maven Daemon - FastClean Maven Plugin</name>
+
+    <properties>
+        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>1.8</maven.compiler.source>
+
+        <roaster.version>2.22.2.Final</roaster.version>
+        <maven.plugin-tools.version>3.6.0</maven.plugin-tools.version>
+        <maven-plugin-plugin.version>${maven.plugin-tools.version}</maven-plugin-plugin.version>
+        <maven.version>3.6.0</maven.version>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.apache.maven</groupId>
+                <artifactId>maven-plugin-api</artifactId>
+                <version>${maven.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.maven.plugin-tools</groupId>
+                <artifactId>maven-plugin-annotations</artifactId>
+                <version>${maven.plugin-tools.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.maven</groupId>
+                <artifactId>maven-core</artifactId>
+                <version>${maven.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-plugin-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven.plugin-tools</groupId>
+            <artifactId>maven-plugin-annotations</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-core</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.forge.roaster</groupId>
+            <artifactId>roaster-jdt</artifactId>
+            <version>${roaster.version}</version>
+        </dependency>
+
+    </dependencies>
+
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-plugin-plugin</artifactId>
+                    <version>${maven-plugin-plugin.version}</version>
+                </plugin>
+            </plugins>
+        </pluginManagement>
+
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-plugin-plugin</artifactId>
+                <configuration>
+                    <goalPrefix>mvnd-build</goalPrefix>
+                    <mojoDependencies>
+                        <dep>org.apache.maven:maven-plugin-api</dep>
+                    </mojoDependencies>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/fastclean/src/main/java/org/mvndaemon/mvnd/plugin/fastclean/FastCleanLifecycleProvider.java
+++ b/fastclean/src/main/java/org/mvndaemon/mvnd/plugin/fastclean/FastCleanLifecycleProvider.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.mvndaemon.mvnd.plugin.fastclean;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+import org.apache.maven.eventspy.AbstractEventSpy;
+import org.apache.maven.lifecycle.DefaultLifecycles;
+import org.apache.maven.lifecycle.mapping.LifecyclePhase;
+import org.slf4j.LoggerFactory;
+
+@Singleton
+@Named
+public final class FastCleanLifecycleProvider extends AbstractEventSpy {
+
+    @Inject
+    public FastCleanLifecycleProvider(DefaultLifecycles defaultLifecycles) {
+        LoggerFactory.getLogger(getClass()).warn("FastCleanLifecycleProvider initialized");
+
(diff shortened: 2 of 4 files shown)


ppalaga:
In addition to the above, the users would have to unbind the stock clean plugin, no?

I really do not think we should add duplicates of existing plugins here in the mvnd project. As I said, the feature should ideally go to the clean plugin where also non-mvnd users could benefit from it.  If that turns out not being doable for some reason, we could create a separate fast clean project under the mvndaemon org. Binding the release cycle of the fast clean plugin to mvnd would unnecessarily complicate the maintenance of the end user projects, because they would have to upgrade the plugin with every new mvnd version.

gnodet:
> In addition to the above, the users would have to unbind the stock clean plugin, no?

No, the default clean plugin is completely override by https://github.com/mvndaemon/mvnd/pull/538/files#diff-627fb1e283acdedd68fabea3467e2bb7c3f6bde8e318d9c7613afba05c881a4eR34-R35

> I really do not think we should add duplicates of existing plugins here in the mvnd project. As I said, the feature should ideally go to the clean plugin where also non-mvnd users could benefit from it. If that turns out not being doable for some reason, we could create a separate fast clean project under the mvndaemon org. Binding the release cycle of the fast clean plugin to mvnd would unnecessarily complicate the maintenance of the end user projects, because they would have to upgrade the plugin with every new mvnd version.

Yes, I somewhat agree.  I wanted to see if the experiment is something worth to investigate if it actually causes a big difference when running on windows.  This could definitely be an improvement to the `clean` plugin.   And fwiw, in this PR, the extension is actually unrelated to `mvnd`, so it should at least be a separate project if we pursue the idea as a standalone extension.

atlib77:
Wow!  You... are... quick!! :)  Thank you so much for taking the time to test out the idea.

So I gave it a shot and strangely, despite you spinning of a new thread, the task didn't finish until the thread completed.  I don't really understand why.  I made some changes to it by using a ExecutorService:

private static ExecutorService executorService = Executors.newSingleThreadExecutor();
and then 
executorService.submit(() -> deleteDir(delDir, retryOnError));

and then it seem to finish the maven-goal right away (and process the delete in the background)

To summarize the results a bit for my specific case (13 projects in build):

**Time spent when using the fastclean plugin:**
clean goal: 174 ms
total build time: 01:51 min (Wall Clock)

**Time spent when using the stock maven clean plugin:**
clean goal: 62263 ms
total build time:  02:15 min (Wall Clock)

So approx. 20% reduction in clean&build time!  I think that's definitely worth something :)  Could possibly be sped up further by delaying the deletes (or lowering priority of it)

(Using the fastclean plugin in mvn with -T1 it took 536 ms)

So I'm really hopeful this could go somewhere.  However, it feels like this has to be run within the maven daemon, as opposed to just maven, in order to be able to run this properly.  Otherwise we'll take the hit during maven shutdown.

Once again, thanks a lot!
Atli


ps.
btw, I also skipped the tmpDir and renamed directly (instead of moving the delDir to the target folder). I did that because if you start a new build while the delete is still ongoing, you get into problems:

                String name = directory.getFileName().toString() + "-" + random;
                Path delDir = directory.resolveSibling(name);
                try {
                    Files.move(directory, delDir);
                    Files.createDirectory(directory);



---

Some additional numbers from my tests:

[INFO] Plugins in lifecycle Phases:
[INFO]
[INFO] clean:
[INFO]      206 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
[INFO]        2 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
[INFO]      470 ms: org.apache.maven.plugins:maven-clean-plugin:3.1.0:clean:default-clean
[INFO]    11531 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
[INFO]     9513 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
[INFO]        2 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
[INFO]    12009 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
[INFO]     1926 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
[INFO]        2 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
[INFO]     2377 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
[INFO]     4097 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
[INFO]     2432 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
[INFO]    17696 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean


**vs fastclean:**

[INFO] Plugins in lifecycle Phases:
[INFO]
[INFO] clean:
[INFO]       23 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]       19 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]        2 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]       12 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]       20 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]       10 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]       15 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]       13 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]       16 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]        9 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]        2 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]       16 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean
[INFO]       17 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean

gnodet:
> Wow! You... are... quick!! :) Thank you so much for taking the time to test out the idea.
> 
> So I gave it a shot and strangely, despite you spinning up a new thread, the maven clean goal didn't finish until the thread completed. I don't really understand why. I made some changes to the code by using a ExecutorService (might be good to serialize the deletion as well):
> 
> `private static ExecutorService executorService = Executors.newSingleThreadExecutor();`
> 
> and then
> 
> `executorService.submit(() -> deleteDir(delDir, retryOnError));`
> 
> and then it seem to finish the maven-goal right away (and process the delete in the background)
> 
> To summarize the results a bit for my specific case (13 projects in build):
> 
> **Time spent when using the fastclean plugin:** clean goal: 174 ms total build time: 01:51 min (Wall Clock)
> 
> **Time spent when using the stock maven clean plugin:** clean goal: 62263 ms total build time: 02:15 min (Wall Clock)
> 
> So approx. 20% reduction in clean&build time! I think that's definitely worth something :) Could possibly be sped up further by delaying the deletes (or lowering priority of it)
> 
> (Using the fastclean plugin in mvn with -T1 it took 536 ms)
> 
> So I'm really hopeful this could go somewhere. However, it feels like this has to be run within the maven daemon, as opposed to just maven, in order to be able to run this properly. Otherwise we'll take the hit during maven shutdown.
> 
> Once again, thanks a lot! Atli
> 
> ps. btw, I also skipped the tmpDir and renamed directly (instead of moving the delDir to the target folder). I did that because if you start a new build while the delete is still ongoing, you get into problems:
> 
> ```
>             String name = directory.getFileName().toString() + "-" + random;
>             Path delDir = directory.resolveSibling(name);
>             try {
>                 Files.move(directory, delDir);
>                 Files.createDirectory(directory);
> ```
> 
> Some additional numbers from my tests:
> 
> [INFO] Plugins in lifecycle Phases: [INFO] [INFO] clean: [INFO] 206 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean [INFO] 2 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean [INFO] 470 ms: org.apache.maven.plugins:maven-clean-plugin:3.1.0:clean:default-clean [INFO] 11531 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean [INFO] 9513 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean [INFO] 2 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean [INFO] 12009 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean [INFO] 1926 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean [INFO] 2 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean [INFO] 2377 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean [INFO] 4097 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean [INFO] 2432 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean [INFO] 17696 ms: org.apache.maven.plugins:maven-clean-plugin:2.5:clean:default-clean
> 
> **vs fastclean:**
> 
> [INFO] Plugins in lifecycle Phases: [INFO] [INFO] clean: [INFO] 23 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 19 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 2 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 12 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 20 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 10 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 15 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 13 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 16 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 9 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 2 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 16 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean [INFO] 17 ms: org.mvndaemon.mvnd:mvnd-fastclean-maven-plugin:0.7.2-SNAPSHOT:clean:default-clean

It would be worth trying with the modified [maven-clean-plugin](https://github.com/apache/maven-clean-plugin/pull/6) .
I've added a join which cause maven to wait until the deletion is finished, but this should only happen at the end of the maven execution.  My assumption is that if it's waiting, you can either kill maven (and only the `[root]/target/.clean` directory will remain), or if you used `mvn clean verify` for example, the deletion should occur in parallel with the actual build.

I do encourage you to continue the discussion in https://github.com/apache/maven-clean-plugin/pull/6, in particular, if you could provide a project where the problem can be reproduced to further test, it would be awesome.

I'm closing this one as I don't see any pushback to improve the `maven-clean-plugin`.

PR #539, opened 2021-12-17
Do not cut sha256 in release workflow

vorburger:
This makes it much easier to actually use the checksum in an automated download script by using `sha256sum -c mvnd*.sha256 mvnd*.zip`.

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -191,9 +191,9 @@ jobs:
 
     - name: Compute sha256 checksums
       run: |
-        sha256sum mvnd-linux-amd64.zip/mvnd-${{ env.VERSION }}-linux-amd64.zip | cut -d ' ' -f 1 > mvnd-linux-amd64.zip/mvnd-${{ env.VERSION }}-linux-amd64.zip.sha256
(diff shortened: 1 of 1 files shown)


vorburger:
> use the checksum in an automated download script

such as e.g. https://github.com/vorburger/vorburger-dotfiles-bin-etc/blob/master/install-github.sh

ecki:
Is this compatible with the maven repository format?

ppalaga (approved):
I was adding the checksums to make some package manager happy. Unfortunately, I do not remember anymore which one of Brew, SDKMAN or Chocolatey was that. Let's try like this.

The CI does not work for some reason. Let me check.

ppalaga:
> Let me check.

I cannot :/ after the repo moved to ASF.

olamy:
> > Let me check.
> 
> I cannot :/ after the repo moved to ASF.

what sort of karma do you need?

gnodet:
> > > Let me check.
> > 
> > 
> > I cannot :/ after the repo moved to ASF.
> 
> what sort of karma do you need?

commitership ? I suppose Peter lost his write access to the code tree during the move given he is not a Maven committer.

olamy:
> > > > Let me check.
> > > 
> > > 
> > > I cannot :/ after the repo moved to ASF.
> > 
> > 
> > what sort of karma do you need?
> 
> commitership ? I suppose Peter lost his write access to the code tree during the move given he is not a Maven committer.

Ah yes makes sense. 
I will try to fix that ;)

aalmiray:
FWIW #574 addresses this with JReleaser

Issue #561, opened 2022-01-04
The same build fails on the second run


suninformation:
The same command is executed successfully for the first time but is abnormal when it is executed again.
Two command execution console outputs are as follows:

```shell
➜  goldenriver-drc git:(master) ✗ mvnd clean compile ymate:entity -DshowOnly=true
[INFO] Processing build on daemon 120445ee
[INFO] Scanning for projects...
[INFO] BuildTimeEventSpy is registered.
[INFO]
[INFO] Using the SmartBuilder implementation with a thread count of 3
[INFO] Task segments : [clean, compile, ymate:entity]
[INFO] Build maximum degree of concurrency is 3
[INFO] Total number of projects is 1
[INFO]
[INFO] ------------------< com.goldenriver:goldenriver-drc >-------------------
[INFO] Building goldenriver-drc 1.0-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO]
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ goldenriver-drc ---
[INFO]
[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ goldenriver-drc ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] Copying 1 resource
[INFO]
[INFO] --- maven-compiler-plugin:2.5.1:compile (default-compile) @ goldenriver-drc ---
[INFO] Compiling 19 source files to /Users/suninformation/IdeaProjects/YMP_2019/goldenriver-drc-parent/goldenriver-drc/target/classes
[INFO]
[INFO] --- ymate-maven-plugin:1.0.0:entity (default-cli) @ goldenriver-drc ---
[INFO] Found and load the config file: /Users/suninformation/IdeaProjects/YMP_2019/goldenriver-drc-parent/goldenriver-drc/src/main/resources/ymp-conf.properties
[INFO]
__  __ __  ___ ___
\ \/ //  |/  // _ \
 \  // /|_/ // ___/
 /_//_/  /_//_/  www.ymate.net
[INFO] Initializing ymate-platform-core-2.1.0-Release build-2022-01-04T16:31:32Z - debug:true - env:dev - PID:50137
[INFO] Initializing ymate-platform-persistence-jdbc-2.1.0-Release build-2022-01-04T16:31:32Z
[INFO] RecycleHelper has registered the number of resources to be recycled: 0
[INFO] Initialization completed, Total time: 63ms
[...]

PR #566, opened 2022-01-05
Remove previous MavenProject class, #561

gnodet:


jglick:
Amends #420?

gnodet:
> Amends #420?

Yes, #523 is a better fix for the concurrency problem than #420, but the custom `MavenProject` was not removed and should have.

PR #573, opened 2022-01-07
Make sure the maven.home and maven.conf properties are correctly set …

### LINKED ISSUE #553: The ${MVND_HOME}/mvn/conf/setting.xml is not used
When I installed it on the windows system and replaced the conf/setting.xml in mvn in the directory with the configuration I used normally, it didn’t seem to take effect, which made me unable to package normally.

After researching, it seems that the configuration can be specified through maven.settings in mvnd.properties. The example given is `~/.m2/settings.xml`.
In other words, I can directly specify my configuration file here, so I wrote this:
```
maven.settings=C:\ProgramData\chocolatey\lib\maven\apache-maven-3.8.3\conf\settings.xml
```

But it still didn't work when I tested it.
For example, when I compile under D:/root, mvnd tells me that the directory of setting.xml is: 
```
D:/root\C:ProgramDatachocolateylibmavenapache-maven-3.8.3confsettings.xml
```

After flipping through, I found this:
https://github.com/apache/maven-mvnd/issues/383
Thanks to @ppalaga for providing an example: /path/to/settings.xml
It made me realize that it may be because the directory symbol "\\" under windows is considered to be an escape character
I tried to modify it
```
maven.settings=C:/ProgramData/chocolatey/lib/maven/apache-maven-3.8.3/conf/settings.xml
```
everything is normal

I recommend using File.separator to determine the file symbol of the current system and replace the configuration

gnodet:
…when the JVM is started, fixes #553

### DIFF
--- a/client/src/main/java/org/mvndaemon/mvnd/client/DaemonConnector.java
+++ b/client/src/main/java/org/mvndaemon/mvnd/client/DaemonConnector.java
@@ -372,6 +372,9 @@ public class DaemonConnector {
             }
 
             Environment.MVND_HOME.addCommandLineOption(args, mvndHome.toString());
+            args.add("-Dmaven.home=" + mvndHome.resolve("mvn"));
+            args.add("-Dmaven.conf=" + mvndHome.resolve("mvn/conf"));
+
             Environment.MVND_JAVA_HOME.addCommandLineOption(args, parameters.javaHome().toString());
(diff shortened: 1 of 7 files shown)


ppalaga (approved):

gnodet:
Need to find out why the tests are failing now...

Issue #560, opened 2022-01-04
Let JVM set max heap size instead of a default value of 2GB (`mvnd.maxHeapSize`)


hashhar:
Have we considered making the `mvnd.maxHeapSize` config default to null so that the JVM can decide the max heap size instead of the current default of 2GB?

This will bring it in line with vanilla Maven which leaves the heap size to the JVM unless an explicit `-Xmx` is passed (either on command line, MAVEN_OPTS or other mechanisms listed on https://maven.apache.org/configure.html).

With current out-of-box behaviour of mvnd it's a bit frustrating to discover that the daemon is crashing because it is low on memory especially since the observable failure is the lost keepalives instead of an OOM.

(See https://github.com/apache/maven-mvnd/issues/291#issuecomment-1003319291 for more context)

cc: @gnodet created as requested.

I'll be happy to submit a patch if we can agree on some behavior here.

ppalaga:
> making the `mvnd.maxHeapSize` config default to null so that the JVM can decide the max heap size instead of the current default of 2GB?

Sounds like a good idea. Thanks for reporting @hashhar!

gnodet:
> Have we considered making the `mvnd.maxHeapSize` config default to null so that the JVM can decide the max heap size instead of the current default of 2GB?
> 
> This will bring it in line with vanilla Maven which leaves the heap size to the JVM unless an explicit `-Xmx` is passed (either on command line, MAVEN_OPTS or other mechanisms listed on https://maven.apache.org/configure.html).
> 
> With current out-of-box behaviour of mvnd it's a bit frustrating to discover that the daemon is crashing because it is low on memory especially since the observable failure is the lost keepalives instead of an OOM.
> 
> (See [#291 (comment)](https://github.com/apache/maven-mvnd/issues/291#issuecomment-1003319291) for more context)
> 
> cc: @gnodet created as requested.
> 
> I'll be happy to submit a patch if we can agree on some behavior here.

@hashhar I think not setting the max heap size sounds reasonable.  Do you want to provide a PR for that ?

hashhar:
Thanks for the decision. I'll work on the PR over this weekend, expect it soon.

cheneychoi:
@hashhar hey bro,i met the same situation ,is this issue solved?

PR #577, opened 2022-01-10
Set default max heap size to null

hashhar:
Let the JVM decide the max heap size instead of using hardcoded defaults
to match the behaviour of vanilla Maven.

Fixes https://github.com/apache/maven-mvnd/issues/560

I tried adding a test like:

```java
Client client = ClientFactory.create(parameters.withJdkJavaOpts("-Xmx200M -XshowSettings"));
final TestClientOutput output = new TestClientOutput();
client.execute(output, "verify").assertSuccess();
output.assertContainsMatchingSubsequence("Max. Heap Size: 200.00M");
```

But it fails. It seems the TestClientOutput only contains messages from Maven itself - not the JVM.

Should I try to add a test that verifies that explicitly passing `mvnd.maxHeapSize` is respected and that explicitly passing `-Xmx` works (instead of getting overwritten by `mvnd.maxHeapSize`).


common/src/main/java/org/mvndaemon/mvnd/common/Environment.java
```diff
@@ -209,7 +209,7 @@
     /**
      * The <code>-Xmx</code> value to pass to the daemon
      */
-    MVND_MAX_HEAP_SIZE("mvnd.maxHeapSize", null, "2G", OptionType.MEMORY_SIZE, Flags.DISCRIMINATING),
+    MVND_MAX_HEAP_SIZE("mvnd.maxHeapSize", null, null, OptionType.MEMORY_SIZE, Flags.DISCRIMINATING | Flags.OPTIONAL),
```
hashhar, line 212: I assumed it's fine to keep the heap size as a discriminating flag still since the heap size cannot be changed on the fly and hence is useful to identify the client within the registry.

gnodet:
> Let the JVM decide the max heap size instead of using hardcoded defaults to match the behaviour of vanilla Maven.
> 
> Fixes #560
> 
> I tried adding a test like:
> 
> ```java
> Client client = ClientFactory.create(parameters.withJdkJavaOpts("-Xmx200M -XshowSettings"));
> final TestClientOutput output = new TestClientOutput();
> client.execute(output, "verify").assertSuccess();
> output.assertContainsMatchingSubsequence("Max. Heap Size: 200.00M");
> ```
> 
> But it fails. It seems the TestClientOutput only contains messages from Maven itself - not the JVM.
> 
> Should I try to add a test that verifies that explicitly passing `mvnd.maxHeapSize` is respected and that explicitly passing `-Xmx` works (instead of getting overwritten by `mvnd.maxHeapSize`).

You should be able to get the JVM args by launching the following command inside the test:
```
mvnd org.codehaus.gmaven:groovy-maven-plugin:2.1.1:execute -Dsource="println java.lang.management.ManagementFactory.getRuntimeMXBean().getInputArguments()"
```
and check that the `-Xmx` string does not appear, or appear with the correct number.

gnodet (changes requested):
Please provide an integration test testing the default and non default values.

oscerd (approved):

hashhar:
Sorry for the delay. I have a working test locally now and I'll push in a while. Thanks for guidance @gnodet

hashhar:
@gnodet Can you PTAL. 2 of the 3 tests are working. I've been unable to figure out how to pass mvnd options/get tests to read mvnd.properties to verify that values from mvnd.properties or on command line are being honored or not.

hashhar:
Ping @gnodet

gnodet:
Superseded by #610

PR #580, opened 2022-01-11
Daemon reuse ignores differences in `.mvn/jvm.config`, fixes #576

### LINKED ISSUE #576: Daemon reuse ignores differences in `.mvn/jvm.config`
I was experimenting with adding the `--add-exports` lines to `.mvn/jvm.config` to make  [Error Prone](https://errorprone.info/docs/installation#maven) work without requiring `<fork>true</fork>`, in order to gain the benefits of the compiler warming up. This seems to work, but `mvnd` reuses a daemon with different JVM flags.

Reproduce:

1. Do a build with a single JVM config like `-Xmx1G`
2. Add another JVM config like `--add-exports=...`
3. Do another build and notice it incorrectly reuses the existing daemon

gnodet:


client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
```diff
@@ -114,6 +115,12 @@ public static void main(String[] argv) throws Exception {
 
         System.setProperty(Environment.MVND_HOME.getProperty(), parameters.mvndHome().toString());
 
+        // .mvn/jvm.config
+        if (Files.isRegularFile(parameters.jvmConfigPath())) {
+            String jvmArgs = Files.lines(parameters.jvmConfigPath()).collect(Collectors.joining(" "));
```
ppalaga, line 120: The stream returned by Files.lines() should be closed.

client/src/main/java/org/mvndaemon/mvnd/client/DefaultClient.java
```diff
@@ -114,6 +115,12 @@ public static void main(String[] argv) throws Exception {
...
+            parameters = parameters.withJvmArgs(" " + jvmArgs);
```
ppalaga, line 121: So we do not care in any way to combine the content of `jvm.config` with `mvnd.jvmArgs`. If the user passes `-Xmx` via both, we pass `-Xmx` twice to `java` executable, possibly with two distinct values. I do not know whether `java` accepts that and whether it accepts the first or the last value. Anyway, it is might become a source of confusion for the users - "I am passing this value via jvm.config but it does not get effective.... Why oh why?!?! Oh, geez, I am also passing it via mvnd.jvmArgs and only that one is accepted!". I wonder whether we could somehow prevent that potential confusion?

gnodet: You can also use `mvnd.maxHeapSize`...
OpenJDK does accept multiple arguments for `-Xmx` and the first one takes precedence (iirc).  We could check the options that we do support explicitely, but we can't validate all JVM options...

ppalaga: > we can't validate all JVM options...

Yes, any code to improve the UX might get complex, hard to maintain and still not covering 100% of cases.

Maybe we should at least mention this combination issue (and the order of precedence) in the help entries of `mvnd.jvmArgs`, `mvnd.maxHeapSize` and add a help entry about `jvm.config`

ppalaga (approved):

Issue #541, opened 2021-12-18
Doesn't seem to work with Quarkus 2.0 dev


sdaschner:
When using `mvnd` with a Quarkus 2.X project, the output looks different to what I'm used to with `mvn`.

More importantly, the keyboard input doesn't work. Is there something I need to configure?

See the two screenshots for a comparison:

mvnd (version `0.7.1-linux-amd64`):

![2021-12-18-150208_1920x1080_scrot](https://user-images.githubusercontent.com/6815170/146643809-14937a97-6810-4002-aa0e-986b4136b21d.png)

mvn:

![2021-12-18-150517_1920x1080_scrot](https://user-images.githubusercontent.com/6815170/146643838-ad7c06d3-ac38-44b2-bb07-a9e70d6d5512.png)

gnodet:
Interactions using the keyboard are currently only supported for maven plugins which use the plexus interaction layer.  The `System.in` input stream is currently not redefined on the daemon, also, the client does read the keyboard and perform specific tasks using the `+`, `-`, `<enter>`, `Ctrl+L` keys, and we'd have to add a mode to switch between local / remote input.  It may be done automatically if we detect the daemon process is trying to read from the input stream, but that's definitely not a trivial thing to implement.

gnodet:
Fwiw, quarkus uses low level terminal interactions to set up the console support, however, there's no terminal in the daemons.  At the OS level, the daemon process input stream is not used at all and the output/error streams are redirected to files.  Inside the JVM, the output/error are conveyed back to the client, and the input is currently not supported.   Supporting redirecting input from the client to the daemon is doable, but setting up a terminal, while in theory doable, would require additional native methods not provided by jansi at the moment (such as `dup2`) and thus not doable right now.

PR #581, opened 2022-01-12
Support redirecting input from client to daemon, #541

gnodet:
The implementation currently switches on the redirection when the daemon actually starts reading the System.in stream using InputStream.read() or InputStream.available().


rmannibucau:
I'm also hitting this issue.

I tested this PR (thanks @gnodet ), it fixes the "blocking" side of the feature but from idea terminal I cant type anything.

PR #585, opened 2022-01-18
Fix user's pronoun in mvnd.properties comment

findepi:


dist/src/main/distro/conf/mvnd.properties
```diff
@@ -86,7 +86,7 @@
 # MVND_MIN_THREADS
 # The minimum number of threads to use when constructing the default {@code -T} parameter for the daemon.
 # This value is ignored if the user passes @{@code -T}, @{@code --threads} or {@code -Dmvnd.threads} on the command
-# line or if he sets {@code mvnd.threads} in {@code ~/.m2/mvnd.properties}.
+# line or if they set {@code mvnd.threads} in {@code ~/.m2/mvnd.properties}.
```
gnodet, line 89: I don't think we can use _they_ and _user_ which is singular.  We could use a passive sentence though: `or if {@code mvnd.threads} is set in ...` which would be better because the value could have been set in the git repo and not directly by the user.

gnodet (changes requested):

dist/src/main/distro/conf/mvnd.properties
Stephan202: The use of _they_ in this context is valid; it's called the [singular they](https://en.wikipedia.org/wiki/Singular_they). (But the passive alternative works too :+1:.)

Stephan202 (approved):

dist/src/main/distro/conf/mvnd.properties
gnodet: I'd rather use the passive one.

findepi: @gnodet @Stephan202 which you prefer?

Stephan202: Right now the text mixes several grammatical forms. How about:
```
# The minimum number of threads to use when constructing the default {@code -T} 
# parameter for the daemon. This value is ignored if @{@code -T}, @{@code
# --threads} or {@code -Dmvnd.threads} is explicitly specified on the command
# line, or if {@code mvnd.threads} is specified in {@code ~/.m2/mvnd.properties}.
```
This avoids any explicit mention of the user.

findepi: Thanks @Stephan202 changed to your suggestion. I omitted, "explicitly" (in "explicitly specified"), as it sounded unnecessary.

gnodet: @Stephan202 can you change the PR with your last proposal please ?

Stephan202: @gnodet I don't have access to this branch; @findepi did apply the change (modulo the word "explicitly", which LGTM).

findepi: I enabled `Allow edits by maintainers` if this is needed.

gnodet (approved):

findepi:
thanks @gnodet for the merge.

PR #591, opened 2022-01-21
Upgrade SLF4J to version 1.7.35

oscerd:
2022-01-13 - Release of SLF4J 1.7.33

• SLF4J now ships with the slf4j-reload4j module delegating to the reload4j backend. Reload4j is a drop-in replacement for log4j version 1.2.17.

• SimpleLogger now prints the thread Id if instructed to do so. This fixes SLF4J-499 requested by Michael Osipov

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
         <logback.version>1.2.10</logback.version>
         <maven.version>3.8.4</maven.version>
         <maven.resolver.version>1.7.2</maven.resolver.version>
-        <slf4j.version>1.7.32</slf4j.version>
+        <slf4j.version>1.7.35</slf4j.version>
 
         <!-- plugin versions a..z -->
         <buildnumber-maven-plugin.version>1.4</buildnumber-maven-plugin.version>


gnodet:
> 2022-01-13 - Release of SLF4J 1.7.33
> 
> • SLF4J now ships with the slf4j-reload4j module delegating to the reload4j backend. Reload4j is a drop-in replacement for log4j version 1.2.17.
> 
> • SimpleLogger now prints the thread Id if instructed to do so. This fixes SLF4J-499 requested by Michael Osipov

FTR, I don't think `mvnd` is impacted by those two issues as the backend used is `logback`.

ppalaga (approved):
> > 2022-01-13 - Release of SLF4J 1.7.33
> > • SLF4J now ships with the slf4j-reload4j module delegating to the reload4j backend. Reload4j is a drop-in replacement for log4j version 1.2.17.
> > • SimpleLogger now prints the thread Id if instructed to do so. This fixes SLF4J-499 requested by Michael Osipov
> 
> FTR, I don't think `mvnd` is impacted by those two issues as the backend used is `logback`.

Yes, we have not used log4j 1.x, but we do use slf4j-simple in the client.

Thanks the PR @oscerd

gnodet:
> > 
> > FTR, I don't think `mvnd` is impacted by those two issues as the backend used is `logback`.
> 
> Yes, we have not used log4j 1.x, but we do use slf4j-simple in the client.
> 

Right, I forgot about that.  That makes we wonder if the following file is used at all:
  https://github.com/apache/maven-mvnd/blob/79137867e51e04e24ae1e924afbd2e4f77e2d860/dist/src/main/distro/conf/logback-client.xml

ppalaga:
> makes we wonder if the following file is used at all:
> https://github.com/apache/maven-mvnd/blob/79137867e51e04e24ae1e924afbd2e4f77e2d860/dist/src/main/distro/conf/logback-client.xml

Yes, it is probably a left-over from earlier times when we used or wanted to use logback on the client.

PR #599, opened 2022-02-02
Adding note on removing quarantine flag on macOS

gunnarmorling:
Hey @ppalaga, a small README addition for manual install on macOS.


ppalaga (approved):
Thanks @gunnarmorling! Does this need to be done also when installing via Brew or Sdkman?

PR #612, opened 2022-03-22
Make `DaemonRegistry.getProcessId0` more robust

### LINKED ISSUE #608: mvnd on Windows throws `java.lang.NumberFormatException: For input string: "self"`
As in #607,

```
mvnd JVM client 0.7.1-windows-amd64 (97c587c11383a67b5bd0ff8388bd94c694b91c1e)
Terminal: org.jline.terminal.impl.jansi.win.JansiWinSysTerminal
Apache Maven 3.8.3 (ff8e977a158738155dc465c6a97ffaf31982d739)
Maven home: C:\Program Files\mvndaemon\mvnd-0.7.1-windows-amd64\mvn
Java version: 11.0.13, vendor: Oracle Corporation, runtime: C:\Program Files\OpenJDK\openjdk-11.0.13_8
Default locale: en_US, platform encoding: Cp1252
OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"
```

does not seem to work to run builds (anything other than `-version`):

```
Exception in thread "main" java.lang.ExceptionInInitializerError
        at org.mvndaemon.mvnd.client.DefaultClient.execute(DefaultClient.java:196)
        at org.mvndaemon.mvnd.client.DefaultClient.main(DefaultClient.java:118)
Caused by: java.lang.NumberFormatException: For input string: "self"
        at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.base/java.lang.Integer.parseInt(Integer.java:652)
        at java.base/java.lang.Integer.parseInt(Integer.java:770)
        at org.mvndaemon.mvnd.common.DaemonRegistry.getProcessId0(DaemonRegistry.java:290)
        at org.mvndaemon.mvnd.common.DaemonRegistry.<clinit>(DaemonRegistry.java:271)
        ... 2 more
```

Seems distinct from #147 / #149.

jglick:
Fixes #608. Tested with Java 11 in non-native mode (from `./dist/target/mvnd-0.7.2-SNAPSHOT-linux-amd64/bin/`) on Linux and Windows 10.

### DIFF
--- a/common/src/main/java/org/mvndaemon/mvnd/common/DaemonRegistry.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/DaemonRegistry.java
@@ -271,23 +271,33 @@ public class DaemonRegistry implements AutoCloseable {
     private static final int PROCESS_ID = getProcessId0();
 
     private static int getProcessId0() {
-        String pid = null;
         try {
             final Path self = Paths.get("/proc/self");
             if (Files.exists(self)) {
-                pid = self.toRealPath().getFileName().toString();
+                String pid = self.toRealPath().getFileName().toString();
+                if (pid.equals("self")) {
+                    LOGGER.debug("/proc/self symlink could not be followed");
+                } else {
+                    LOGGER.debug("loading own PID from /proc/self link: {}", pid);
+                    try {
+                        return Integer.parseInt(pid);
+                    } catch (NumberFormatException x) {
+                        LOGGER.warn("Unable to determine PID from malformed /proc/self link `" + pid + "`");
+                    }
+                }
             }
         } catch (IOException ignored) {
+            LOGGER.debug("could not load /proc/self", ignored);
         }
-        if (pid == null) {
-            pid = ManagementFactory.getRuntimeMXBean().getName().split("@", 0)[0];
-        }
-        if (pid == null) {
-            int rpid = new Random().nextInt(1 << 16);
-            LOGGER.warn("Unable to determine PID, picked a random number=" + rpid);
-            return rpid;
-        } else {
+        String vmname = ManagementFactory.getRuntimeMXBean().getName();
+        String pid = vmname.split("@", 0)[0];
+        LOGGER.debug("loading own PID from VM name: {}", pid);
+        try {
             return Integer.parseInt(pid);
+        } catch (NumberFormatException x) {
+            int rpid = new Random().nextInt(1 << 16);
+            LOGGER.warn("Unable to determine PID from malformed VM name `" + vmname + "`, picked a random number=" + rpid);
+            return rpid;
         }
     }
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/mvndaemon/mvnd/common/DaemonRegistry.java
```diff
@@ -271,23 +271,33 @@ private void reset() {
...
+                    }
+                }
             }
         } catch (IOException ignored) {
+            LOGGER.debug("could not load /proc/self", ignored);
         }
-        if (pid == null) {
-            pid = ManagementFactory.getRuntimeMXBean().getName().split("@", 0)[0];
-        }
-        if (pid == null) {
```
jglick, line 285: AFAICT this branch was impossible previously (`pid` would have been set above), so https://github.com/apache/maven-mvnd/blob/3b100839a810ce9ae2555c1592b5edb623223924/common/src/main/java/org/mvndaemon/mvnd/common/DaemonRegistry.java#L286-L288 dating back to e3b58e80e5e9afca3dc84f137bbac70d6f2edcc5 was dead code?

common/src/main/java/org/mvndaemon/mvnd/common/DaemonRegistry.java
```diff
@@ -271,23 +271,33 @@ private void reset() {
     private static final int PROCESS_ID = getProcessId0();
 
     private static int getProcessId0() {
-        String pid = null;
         try {
             final Path self = Paths.get("/proc/self");
             if (Files.exists(self)) {
```
ppalaga, line 276: I wonder whether we should look into /proc/self only on non-windows and non-mac systems?
As your situation has shown, /proc/self may exists on Windows or Mac. Accessing it not only makes no sense, it also may fail for various reasons, such as insufficient permissions.
So I'd vote for enclosing this in a non-windows and non-mac condition based on `org.mvndaemon.mvnd.common.Os.current()` and using `ManagementFactory.getRuntimeMXBean().getName()` otherwise.

common/src/main/java/org/mvndaemon/mvnd/common/DaemonRegistry.java
```diff
@@ -271,23 +271,35 @@ private void reset() {
     private static final int PROCESS_ID = getProcessId0();
 
     private static int getProcessId0() {
-        String pid = null;
-        try {
-            final Path self = Paths.get("/proc/self");
-            if (Files.exists(self)) {
-                pid = self.toRealPath().getFileName().toString();
+        if (Os.current() == Os.LINUX) {
```
ppalaga, line 274: I originally thought `if (Os.current() != Os.WINDOWS && Os.current() != Os.MAC)` would be better, because both UNKNOWN or any future expansion of the Os enum are likely to be Unix flavors implementing procfs. But I admit that assumption is rather weak and your proposal isn't actually bad because if `ManagementFactory.getRuntimeMXBean().getName().split()` does not work for some UNKNOWN or future Os, then somebody would complain and we can make the condition more specific. Thus :+1:

ppalaga (approved):

ppalaga:
Thanks for your contribution, @jglick!

PR #620, opened 2022-04-25
Reapply Refactor build and release workflows apache#574

gnodet:
This reverts commit e8e5ffc4eeb179ba6aecb0382d9f70697f7d42f7.


gnodet:
[aalmiray](https://github.com/aalmiray) is there a way to publish snapshots to the official apache snapshot repository or do I have to file a request to ASF infra to allow the usage of the GitHub token ?

aalmiray:
It depends. How are those binaries normally published?

gnodet:
> It depends. How are those binaries normally published?

Apache snapshots are usually published using a simple `mvn deploy` which uploads them to the apache snapshot repository at http://repository.apache.org/snapshots/.

Fwiw, I've raised an infra ticket at https://issues.apache.org/jira/browse/INFRA-23200.

aalmiray:
> Apache snapshots are usually published using a simple `mvn deploy` which uploads them to the apache snapshot repository at http://repository.apache.org/snapshots/.

Ah, I see. That would be the case with JARs and POMs which will continue to work using the standard Maven publication plugins. For ZIPs and TARs you'd do the same. 

FWIW the `artifactory` uploader provided by JReleaser may be used to publish artifacts (ZIPs in mvnd's case) to a Nexus instance.

PR #625, opened 2022-04-30
only publish sha256, skip md5+sha1+sha512

hboutemy:
ASF does not need md5 nor sha1 nor sha256: only .asc and .sha512


gnodet:
Homebrew packager will need the sha256, so I'd rather keep them instead of having to regenerate them at a later phase.

gnodet:
> Homebrew packager will need the sha256, so I'd rather keep them instead of having to regenerate them at a later phase.
> See https://github.com/mvndaemon/homebrew-mvnd/blob/master/Formula/mvnd.rb

@hboutemy no problem with that ?

hboutemy:
yes, sha256 is ok regarding ASF policy, then ok for me

PR #626, opened 2022-05-01
prepare Reproducible Builds

hboutemy:
not sure this is sufficient, but at least this is required for Reproducible Builds...


hboutemy:
removing platform-dependant data from daemon common breaks tests...
I'm convinced that such things should not be at that place: perhaps only in client, because client is the only aspect that is native at the end?

ppalaga (approved):
Looks good, thanks @hboutemy !

gnodet:
#630 should help with the `build.properties` file.

PR #635, opened 2022-05-03
Use "official" sisu-maven-plugin

cstamas:
Instead to pull in takari-lifecycle only to use one
single mojo out of it (the index).

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -90,16 +90,8 @@
                 </executions>
             </plugin>
             <plugin>
-                <groupId>io.takari.maven.plugins</groupId>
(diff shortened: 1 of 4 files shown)


gnodet:
Actually, I don't think using the plugin is a good idea.  I'd rather use the situ-inject APT processor which should kick in by simply adding `sisu-inject` to the class path at compile scope.

gnodet (changes requested):
Actually, I don't think using the plugin is a good idea. I'd rather use the situ-inject APT processor which should kick in by simply adding sisu-inject to the class path at compile scope.

PR #652, opened 2022-05-16
Add asdf install method

mattnelson:


gnodet:
I think we'll need the following things in order to properly support a new packager:
  * update the scripts wrt to the new location of published binaries:   https://downloads.apache.org/maven/mvnd/0.8.0/
  * maybe move the https://github.com/joschi/asdf-mvnd repo inside https://github.com/mvndaemon, @joschi 
  * provide an update script so that releases can be update easily, similar to https://github.com/apache/maven-mvnd/blob/master/build/publish-on-homebrew.sh

joschi:
> maybe move the [joschi/asdf-mvnd](https://github.com/joschi/asdf-mvnd) repo inside [@mvndaemon](https://github.com/mvndaemon)

Happy to migrate the repository, but also happy to just keep it in my GitHub account.

> * provide an update script so that releases can be update easily,

@gnodet That's the beauty of asdf-vm: You don't need to update the plugin with every release. You only need to update it when your release/download location changed. 😄

gnodet:
> > update the scripts wrt to the new location of published binaries: [downloads.apache.org/maven/mvnd/0.8.0](https://downloads.apache.org/maven/mvnd/0.8.0/)
> 
> @gnodet Any reason you're not updating the GitHub releases anymore (additionally to the Apache website)?

Because this is the official channel for ASF releases: https://infra.apache.org/release-distribution

joschi (approved):

PR #654, opened 2022-05-19
Make mvnd coloring more consistent with maven, fixes #653

### LINKED ISSUE #653: mvnd goal execution id display inconsistent with Maven
in mvnd, a running mojo is displayed as "plugin:goal @ executionid":
```
:maven-compat            maven-surefire-plugin:test @ default-test
```

but in Maven, @ syntax is for artifactid, execution id is in parenthesis:
```
[INFO] --- maven-assembly-plugin:3.3.0:single (create-distro-packages) @ apache-maven ---
```

such inconsistency will create users confusion (particularly given mvnd also uses ANSI color for mvn's @ artifactId), it would be better to align mvnd with mvn (syntax and ANSI color):
```
:maven-compat            maven-surefire-plugin:test (default-test)
```

and eventually, the current module artifact id could use the same ANSI color as mvn's "@ artifactId"

gnodet:


gnodet:
image

ppalaga:
Thanks, for the images. 
I'd vote for not rendering the execution in bold. I do not think it is more important than other things in the UI and hence the highlighting is not necessary. WDYT @hboutemy ?

ppalaga:
Thanks, for the images. I'd vote for not rendering the execution in bold. I do not think it is more important than other things in the UI and hence the highlighting is not necessary. WDYT @hboutemy ?

ppalaga:
Ups, sorry, I have hit a wrong button.

Issue #662, opened 2022-07-01
Missing argument for option -D


walnut-tom:
$**mvn -D aaa=aaa -version**
Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
Maven home: /usr/local/Cellar/maven/3.8.6/libexec
Java version: 1.8.0_312, vendor: Amazon.com Inc., runtime: /Users/user/Library/Java/JavaVirtualMachines/corretto-1.8.0_312/Contents/Home/jre
Default locale: zh_CN, platform encoding: UTF-8
OS name: "mac os x", version: "12.4", arch: "x86_64", family: "mac"

$**mvnd -D aaa=aaa -version**
Exception in thread "main" java.lang.IllegalArgumentException: Missing argument for option -D
	at org.mvndaemon.mvnd.client.DefaultClient.setSystemPropertiesFromCommandLine(DefaultClient.java:146)
	at org.mvndaemon.mvnd.client.DefaultClient.main(DefaultClient.java:106)

nandorholozsnyak:
Same here, but I'm glad there is already an open issue for that with a "fix" too :)

cstamas:
True with `-D aaa=aaa` but works OK without space, with `-Daaa=aaa`

nandorholozsnyak:
Yes true it works without space with `-D`, and it works with this format too `--define=property=value` but if it really wants to support the Maven based properties then it should be working with the space too. :D

walnut-tom:
use mvn command, I never used the form without space actually too.

but I use **org.apache.maven.shared:maven-invoker:3.2.0** invoke **maven/mvnd** like this:

```java
public class InvokerListClass {

    public static void main(String[] args) throws MavenInvocationException {
        Invoker invoker = new DefaultInvoker();
        invoker.setMavenHome(new File("/Users/xxx/maven-mvnd-0.8.0-darwin-amd64"));
        invoker.setMavenExecutable(new File("mvnd"));
        final InvocationRequest invocationRequest = new DefaultInvocationRequest();
        Properties properties = new Properties();
        properties.setProperty("artifact", "com.h2database:h2:1.4.200");
        invocationRequest.setProperties(properties);
        invocationRequest.setGoals(Collections.singletonList("org.apache.maven.plugins:maven-dependency-plugin:3.2.0:list-classes"));
        invoker.execute(invocationRequest);
    }
}

```

PR #663, opened 2022-07-02
Missing argument for option -D #662

walnut-tom:


gnodet:
@walnut-tom Would it be possible to set up a test case ?

gnodet:
Fixed

PR #680, opened 2022-08-30
System properties should have precedence over environment variables, fixes #675

### LINKED ISSUE #675: How do I change the JDK dynamically
### For example，My environment variables are configured with JDK 17
image

### At this point I need to change the JDK to 8 to compile the project，Because my modification of mvnd.properties does not take effect
image

### I have to modify the environment configuration for this to take effect
image

gnodet:


gnodet:
@ppalaga do you recall why we chose to prioritise environment variables over system properties originally ?

PR #686, opened 2022-09-10
Fixes module id coloring on Maven transfer.

cstamas:
The module id coloring was not consistent during transfer and it caused "vibrating" effect.

https://github.com/apache/maven-mvnd/issues/685

### DIFF
--- a/common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
@@ -90,6 +90,7 @@ public class TerminalOutput implements ClientOutput {
 
     private static final AttributedStyle GREEN_FOREGROUND = new AttributedStyle().foreground(AttributedStyle.GREEN);
     private static final AttributedStyle CYAN_FOREGROUND = new AttributedStyle().foreground(AttributedStyle.CYAN);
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
```diff
@@ -767,12 +768,15 @@ private void addProjectLine(final List<AttributedString> lines, Project prj) {
...
+                    .style(CYAN_FOREGROUND)
                     .append(String.format(artifactIdFormat, prj.id))
+                    .style(WHITE_FOREGROUND)
                     .append(transfer);
         } else if (execution == null) {
             asb
                     .append(':')
-                    .append(prj.id);
+                    .style(CYAN_FOREGROUND)
+                    .append(String.format(artifactIdFormat, prj.id));
```
gnodet, line 779: I wonder if the project id really needs formatting here.  The main purpose was to trim it to a certain length to keep a nice display.  In this case, there's nothing more on the line though...

gnodet (changes requested):

common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
cstamas: fixed

PR #712, opened 2022-10-14
Add 'mvnd.rawStreams' property

### LINKED ISSUE #708: exec-maven-plugin (exec:exec) output unexpected prefix for each line of stdout/stderr
```bash
$ mvnd -B -Dmvnd.noBuffering=true exec:exec -Dexec.executable=uname
[INFO] [stdout] Darwin

$ mvnd -B -Dmvnd.noBuffering=true exec:exec -Dexec.executable=java -Dexec.args=-version
[WARN] [stderr] openjdk version "11.0.2" 2019-01-15
[WARN] [stderr] OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
[WARN] [stderr] OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)

$ mvn -B -Dmvnd.noBuffering=true exec:exec -Dexec.executable=uname
Darwin
```

It is reproducible on mvnd 0.8.2. The output difference breaks the script to switch to mvnd.

gzm55:
1. add `mvnd.rawStreams=<boolean>` property to control the stdout/stderr stream
2. `--raw-streams` cli option overrides `-Dmvnd.rawStreams=...`
3. move `--raw-streams` a mvnd specific option

closes #708


gnodet (changes requested):
Is there any good reason to keep both an option and a property ? I think they should be merged together.

gzm55:
> Is there any good reason to keep both an option and a property ? I think they should be merged together.

if merged, with type `OptionType.VOID`, the [`getCommandLineOption`](https://github.com/apache/maven-mvnd/blob/5e59c404535ba21b925a372340c7b11749aa4050/common/src/main/java/org/mvndaemon/mvnd/common/Environment.java#L425) will ignore the value following `-Dmvnd.rawStreams=....`

gnodet:
> > Is there any good reason to keep both an option and a property ? I think they should be merged together.
> 
> if merged, with type `OptionType.VOID`, the [`getCommandLineOption`](https://github.com/apache/maven-mvnd/blob/5e59c404535ba21b925a372340c7b11749aa4050/common/src/main/java/org/mvndaemon/mvnd/common/Environment.java#L425) will ignore the value following `-Dmvnd.rawStreams=....`. Another property `mvnd.serial` should have the same problem.

I think that's a bug, they should be `OptionType.BOOLEAN`, shouldn't they ?

gzm55:
> > > Is there any good reason to keep both an option and a property ? I think they should be merged together.
> > 
> > 
> > if merged, with type `OptionType.VOID`, the [`getCommandLineOption`](https://github.com/apache/maven-mvnd/blob/5e59c404535ba21b925a372340c7b11749aa4050/common/src/main/java/org/mvndaemon/mvnd/common/Environment.java#L425) will ignore the value following `-Dmvnd.rawStreams=....`. Another property `mvnd.serial` should have the same problem.
> 
> I think that's a bug, they should be `OptionType.BOOLEAN`, shouldn't they ?

For non-parameter switch `-1`, `--serial` or `--raw-streams`, the type should be `VOID`, whereas all the property should not be `VOID`. To support the merging of a property and a void cli switch, we have to assign two types to one `Environment`. I think there should be better another pr to improve these cases.

gnodet:
> > > > Is there any good reason to keep both an option and a property ? I think they should be merged together.
> > > 
> > > 
> > > if merged, with type `OptionType.VOID`, the [`getCommandLineOption`](https://github.com/apache/maven-mvnd/blob/5e59c404535ba21b925a372340c7b11749aa4050/common/src/main/java/org/mvndaemon/mvnd/common/Environment.java#L425) will ignore the value following `-Dmvnd.rawStreams=....`. Another property `mvnd.serial` should have the same problem.
> > 
> > 
> > I think that's a bug, they should be `OptionType.BOOLEAN`, shouldn't they ?
> 
> For non-parameter switch `-1`, `--serial` or `--raw-streams`, the type should be `VOID`, whereas all the property should not be `VOID`. To support the merging of a property and a void cli switch, we have to assign two types to one `Environment`. I think there should be better another pr to improve these cases.

Agreed !

gzm55:
another method, remove `--raw-streams` option, then use `-Dmvnd.rawStreams` as a `void` cli option.

Pros: do not change structure of `Environment`; less incompatible options between `mvnd` and `maven`
Cons: break the cli interface.

gzm55:
prefer #719

PR #716, opened 2022-10-15
Calculate java home from java command

### LINKED ISSUE #715: Calculate java home for native mvnd from java executable in $PATH
To have a more consistent behavior as pure java version, the native `mvnd` should has a fall-back java home when we can search a `java` executable from `$PATH`.

gzm55:
closes #715

### DIFF
--- a/common/src/main/java/org/mvndaemon/mvnd/common/OsUtils.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/OsUtils.java
@@ -16,7 +16,6 @@
 package org.mvndaemon.mvnd.common;
 
 import java.io.BufferedReader;
-import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/mvndaemon/mvnd/common/OsUtils.java
```diff
@@ -101,6 +102,29 @@ public static long findProcessRssInKb(long pid) {
         }
     }
 
+    public static String findJavaHomeFromPATH() {
+        final String basename = Os.current().isUnixLike() ? "java" : "java.exe";
+        final String PATH = System.getenv("PATH");
```
gnodet: Please use lowercase `path` for variable names.

common/src/main/java/org/mvndaemon/mvnd/common/OsUtils.java
```diff
@@ -101,6 +102,29 @@ public static long findProcessRssInKb(long pid) {
...
+        if (null == PATH) {
+            return null;
+        }
+        for (final String dirname : PATH.split(File.pathSeparator)) {
```
gnodet: This is not completely right.  If one of the path element contains an escaped reserved character such as `:`, this won't work.  The split should have to take that into account, maybe by using a regex that ensures that the `pathSeparator` is not preceded by an odd number of backslashes.

common/src/main/java/org/mvndaemon/mvnd/common/OsUtils.java
```diff
@@ -101,6 +102,29 @@ public static long findProcessRssInKb(long pid) {
...
+            final File java = new File(dirname, basename);
```
gnodet: Use nio2 `Path` as in most mvnd code.

common/src/main/java/org/mvndaemon/mvnd/common/OsUtils.java
```diff
@@ -101,6 +102,29 @@ public static long findProcessRssInKb(long pid) {
...
+            if (java.isFile() && java.canExecute()) {
+                final String[] cmd = { java.getAbsolutePath(), "-XshowSettings:properties", "-version" };
+                final List<String> output = new ArrayList<String>(1);
```
gnodet: The size of `1` does not make much sense here.  The output will most probably include much more lines.

common/src/main/java/org/mvndaemon/mvnd/common/OsUtils.java
```diff
@@ -101,6 +102,29 @@ public static long findProcessRssInKb(long pid) {
...
+                exec(cmd, output);
+                final List<String> javaHomeLines = output.stream().filter(l -> l.contains(" java.home = "))
```
gnodet: No need for `final` here.

gnodet (changes requested):

common/src/main/java/org/mvndaemon/mvnd/common/OsUtils.java
gzm55: In the latest commit, the env PATH is not parsed, and the `java` is auto searched by `ProcessBuilder`.

common/src/main/java/org/mvndaemon/mvnd/common/OsUtils.java
gzm55: In the latest commit, `File` is removed.

gnodet: Ah, I missed this commit, thx !

gnodet (approved):

PR #717, opened 2022-10-16
Try native image then fallback to pure java version

gzm55:
The new entry script does some light weight checking to select the first available one from:

- native image
- pure java version

To control the fallback logic, a new environment switch `MVND_CLIENT` is added. Default `auto` enables the fallback logic, set to `native` or `jvm` to force select the native image or pure java version.

Checking of native image is mainly based on the file name, and on Linux, there is another `ldd` check to ensure glibc version is OK.


gnodet:
I like the fallback idea, though I wonder if we should move that fallback code directly into the `mvnd.sh` script, which would delegate to the native binary if possible, eventually renaming `mvnd` to `mvnd-native`.
Also, we may want to investigate a way to run the client on a different JVM that the daemon.  We could add two  mvnd environment variables, one which would point to the default client JVM, and another that would define whether we want to _always_ use that JVM for the client, _auto_ switch to that JVM is the environment one is that JDK 11 compatible, or _never_ use it.  The embedded maven would then only be used if the former is not defined or if the later is set to _never_.

gzm55:
A more complex fallback strategy sounds OK, we could define the rules clearly. And the main entrypoint well known is now `$MVND_HOME/bin/mvnd`. To keep the compatibility, can we move the `mvnd.sh` (with fallback function) to `bin/mvnd` and `bin/mvnd.cmd`, and the native binary are moved to `mvnd-native`/`mvnd-native.exe`?

gnodet:
> A more complex fallback strategy sounds OK, we could define the rules clearly. And the main entrypoint well known is now `$MVND_HOME/bin/mvnd`. To keep the compatibility, can we move the `mvnd.sh`/`mvnd.cmd` (with fallback function) to `mvnd` or `mvnd.cmd`, and the native binary are moved to `mvnd-native`/`mvnd-native.exe`?

Sounds good !

gzm55:
> Also, we may want to investigate a way to run the client on a different JVM that the daemon. We could add two mvnd environment variables, one which would point to the default client JVM, and another that would define whether we want to _always_ use that JVM for the client, _auto_ switch to that JVM is the environment one is that JDK 11 compatible, or _never_ use it. The embedded maven would then only be used if the former is not defined or if the later is set to _never_.

@gnodet do you mean add two new env: `MVND_CLIENT_JAVA_HOME` and `MVND_CLIENT_JAVA_HOME_REQUIRE`, and when the native version fails, we

-  (MVND_CLIENT_JAVA_HOME is unset) or (MVND_CLIENT_JAVA_HOME_REQUIRE=never):
   - JAVA_HOME/bin/java or java is JDK 11+: mvnd with os java
   - JAVA_HOME/bin/java or java is old JDK: embeded maven with os java
-  MVND_CLIENT_JAVA_HOME is set, and MVND_CLIENT_JAVA_HOME_REQUIRE=always
   - always run mvnd with MVND_CLIENT_JAVA_HOME/bin/java
-  MVND_CLIENT_JAVA_HOME is set, and MVND_CLIENT_JAVA_HOME_REQUIRE=auto (default value)
   - MVND_CLIENT_JAVA_HOME/bin/java is JDK 11+: mvnd with MVND_CLIENT_JAVA_HOME/bin/java
   - JAVA_HOME/bin/java or java is JDK 11+: mvnd with os java
   - JAVA_HOME/bin/java or java is old JDK: embeded maven with os java

ppalaga (changes requested):
I'd vote for evaluating the performance impact of wrapping `mvnd` binary inside a shell script before merging this.

gzm55:
> I'd vote for evaluating the performance impact of wrapping `mvnd` binary inside a shell script before merging this.

when change the native image names, the checking step of native can be replaced by a `test -x mvnd-native-<os_arch>`, this would has no performance issue.

ppalaga:
I wonder what is your goal here, @gzm55 ? You want to have single package installable and runnable on any host of any arch and OS combination?

gzm55:
> I wonder what is your goal here, @gzm55 ? You want to have single package installable and runnable on any host of any arch and OS combination?

Goal:
- single official way to startup `mvnd` without knowing the internal structure. now there is only 4 prebuild binaries which depend on hardware and glibc 2.14, and the users have to switch to another script if he fails at first time.
- treat `mvnd` as a `maven` distribution which runs faster than or at least same as, the original maven, then the user do not need to install another maven.

ppalaga:
> Goal:
> 
> * single official way to startup `mvnd` without knowing the internal structure. now there is only 4 prebuild binaries which depend on hardware and glibc 2.14, and the users have to switch to another script if he fails at first time.
> * treat `mvnd` as a `maven` distribution which runs faster than or at least same as, the original maven, then the user do not need to install another maven.

I'd be fine with all of that as long as installing the native binary as the entry point stays possible. For performance reasons package managers like SDKMAN should still prefer doing that if a native client is available for the given platform. What do others think?

gzm55:
> I'd be fine with all of that as long as installing the native binary as the entry point stays possible. For performance reasons package managers like SDKMAN should still prefer doing that if a native client is available for the given platform. What do others think?

we can use a env flag to force the entry script always use the native binary

ppalaga:
Hm... am I the only one thinking that the `mvnd` native executable should be the entry point (rather than a wrapper shell script) on platforms for which we have `mvnd` native executable?

ppalaga:
These changes seem to be addressing what I'd call install time concerns. Whether the given build of mvnd native executable can run on the current machine can be decided once at install time. Why do we need to run all those checks anew on every invocation of mvnd?

gnodet:
> These changes seem to be addressing what I'd call install time concerns. Whether the given build of mvnd native executable can run on the current machine can be decided once at install time. Why do we need to run all those checks anew on every invocation of mvnd?

Because we don't have any installer... ? In particular, people often just unzip the tarball / zip.
  https://github.com/apache/camel/blob/main/.github/actions/install-mvnd/action.yml
An alternative would be to provide an install script from a known location that would download the latest version, unzip and do the checks maybe ...

gzm55:
> These changes seem to be addressing what I'd call install time concerns. Whether the given build of mvnd native executable can run on the current machine can be decided once at install time. Why do we need to run all those checks anew on every invocation of mvnd?

how about we do a persist optimization at the real installing time for each package manager? almost every package manager has a post installing step, in which we can check via `mvnd-native --status`, if passed, then `mv -f mvnd-native mvnd` to overwrite the generic entry script.

ppalaga:
I think both proposals boil down to having a reusable piece of shell/PowerShell for linking/moving `mvnd` to the right file. I like that idea. @gnodet variant would also be able to download the zip.

gzm55:
@ppalaga i add a post-install script `bin/mvnd-persist-native` for detecting and moving the native mage as the default entry

gzm55:
> I'd vote for evaluating the performance impact of wrapping `mvnd` binary inside a shell script before merging this.

@ppalaga I did some basic profiling, comparing to directly execute native binaries:
- `MVND_ENTRY_FALLBACK=false`: fallback script increases about **50ms** on linux and macos
- `MVND_ENTRY_FALLBACK=true`: fallback script increases about **60ms** on macos
- `MVND_ENTRY_FALLBACK=true`: fallback script increases about **120ms** on linux

ppalaga:
I'd dare to vote against making `mvnd` to point at a shell script by default. It is adding indirection and performance penalties on major platforms for which we have proper native executables, thus worsening both the user experience for the majority of users and reliable testability of the default setup.

I'd be open to accepting the following alternatives that I think also satisfy the original original requirement of having a universal single entry mvnd:

1. Enhance the existing shell scripts to fall back to native executables (if available and working on the current platform) without renaming them.
2. Provide install scripts that replace the native mvnd with a shell script if the native mvnd does not work on or does not match the current platform.

What do others think?

gzm55:
> 1. Enhance the existing shell scripts to fall back to native executables (if available and working on the current platform) without renaming them.
> 2. Provide install scripts that replace the native mvnd with a shell script if the native mvnd does not work on or does not match the current platform.

Hi @ppalaga , I would like to make this pr focus on the 1st point, implementing a proper fallback logic in the script entrypoint of the current file layout (`mvnd.sh` and `mvnd.cmd`). And create another issue/pr for the 2nd point to discuss and dev the post installing process.

gzm55:
@ppalaga @gnodet 

> 1. Enhance the existing shell scripts to fall back to native executables (if available and working on the current platform) without renaming them.

done in this pr

> 2. Provide install scripts that replace the native mvnd with a shell script if the native mvnd does not work on or does not match the current platform.

record in the issue #771

dist/src/main/distro/bin/mvnd.cmd
```diff
@@ -25,6 +25,9 @@
 @REM   MAVEN_BATCH_PAUSE (Optional) set to 'on' to wait for a key stroke before ending.
 @REM   MAVEN_OPTS        (Optional) Java runtime options used when Maven is executed.
 @REM   MAVEN_SKIP_RC     (Optional) Flag to disable loading of mavenrc files.
+@REM   MVND_ENTRY_FALLBACK (Optional) Flag to disable fallback to pure java mvnd,
+@REM                       default 'true' enable the fallback,
+@REM                       set to 'false' to force execute the native mvnd.
```
ppalaga, line 30: I wonder if we want to give users a bit more control here. I mean with the proposed implementation users have no way to force the JVM client if they are on natively-supported platform. This might come in handy when debugging. Could we perhaps make it 3-valued, something like 

```
@REM   MVND_CLIENT (Optional) Which mvnd client should be used to communicate with the daemon process.
@REM                       'auto' (default) - prefer the native client mvnd.exe if it is available for the current 
@REM                                                platform and has all dependencies satisfied on the current machine; 
@REM                                                otherwise use the pure Java client
@REM                       'native' - use the native client mvnd.exe
@REM                       'jvm' - use the pure Java client
```

gzm55: new commit switch to the `MVND_CLIENT` with three values.

dist/src/main/distro/bin/mvnd.cmd
```diff
@@ -25,6 +25,11 @@
 @REM   MAVEN_BATCH_PAUSE (Optional) set to 'on' to wait for a key stroke before ending.
 @REM   MAVEN_OPTS        (Optional) Java runtime options used when Maven is executed.
 @REM   MAVEN_SKIP_RC     (Optional) Flag to disable loading of mavenrc files.
+@REM   MVND_CLIENT       (Optional) Control how to select mvnd client to communicate with the daemon:
+@REM                        'auto' (default) - prefer the native client mvnd if it works; otherwise use
+@REM                                           the pure Java client.
+@REM                        'native' - use the native client mvnd.exe
```
ppalaga, line 31: This is windows, so we can use `mvnd.exe` every where, no?

`if it works` sounds a bit unspecific to me. Could we please somehow concisely describe what the script actually does?

```suggestion
@REM   MVND_CLIENT       (Optional) Control how to select mvnd client to communicate with the daemon:
@REM                        'auto' (default) - prefer the native client mvnd.exe if it suits the current OS and processor architecture; otherwise use
@REM                                           the pure Java client.
@REM                        'native' - use the native client mvnd.exe
```

dist/src/main/distro/bin/mvnd.sh
```diff
@@ -25,6 +25,11 @@
 #   JAVA_HOME       Must point at your Java Development Kit installation.
 #   MAVEN_OPTS      (Optional) Java runtime options used when Maven is executed.
 #   MAVEN_SKIP_RC   (Optional) Flag to disable loading of mavenrc files.
+#   MVND_CLIENT     (Optional) Control how to select mvnd client to communicate with the daemon:
+#                      'auto' (default) - prefer the native client mvnd if it works; otherwise use
+#                                         the pure Java client.
+#                      'native' - use the native client mvnd.exe
```
ppalaga, line 31: Similar like above:
```suggestion
#   MVND_CLIENT     (Optional) Control how to select mvnd client to communicate with the daemon:
#                      'auto' (default) - prefer the native client mvnd if it suits the current OS and processor architecture; otherwise use
#                                         the pure Java client.
#                      'native' - use the native client mvnd
```

ppalaga (commented):
Thanks looks great! I left some minor comment text suggestions inline.

dist/src/main/distro/bin/mvnd.cmd
gzm55: OK~

ppalaga (approved):
Great work, thanks a lot, @gzm55!

dist/src/main/provisio/maven-distro.xml
```diff
@@ -97,6 +97,7 @@
             <include>mvnd</include>
             <include>mvnd.exe</include>
         </directory>
+        <file touch="platform-${os.detected.name}-${os.detected.arch}"/>
```
gnodet, line 100: What is the purpose of this line ?

PR #719, opened 2022-10-16
Add a mvnd.rawStreams property

gnodet:


gnodet:
Alternative proposal to #712.
The only drawback is that you can't specify `-Dmvnd.rawStreams=false` but only `-Dmvnd.rawStreams` which sets to value to `true`.

gzm55:
Hi @gnodet I think this pr works fine, treating `-Dmvnd.rawStreams` as a zero parameter switch like `--raw-streams` for now, and i would close #712.
In the mean while, should we leave another issue to record some properties like `mvnd.rawStreams` or `mvnd.serial` can not accept `false` value.

gnodet:
> Hi @gnodet I think this pr works fine, treating `-Dmvnd.rawStreams` as a zero parameter switch like `--raw-streams` for now, and i would close #712.

Good, I just need to fix the failing tests.

> In the mean while, should we leave another issue to record some properties like `mvnd.rawStreams` or `mvnd.serial` can not accept `false` value.

Yes, please do.

PR #721, opened 2022-10-18
Fixup #716 Calculate java home from java command

ppalaga:
Fixup #715

* Add a test
* Simplify the stream processing


gnodet (approved):

client/src/main/java/org/mvndaemon/mvnd/client/DaemonParameters.java
```diff
@@ -150,8 +150,11 @@ public Path javaHome() {
         }
     }
 
-    private String javaHomeFromPath() {
-        final String jHome = OsUtils.findJavaHomeFromPath();
+    private static String javaHomeFromPath() {
+        LOG.warn("Falling back to finding JAVA_HOME by running java executable available in PATH."
```
gzm55, line 154: why warn here? the generic java version (mvnd.sh) will also search java from path without a warning

ppalaga: > why warn here? the generic java version (mvnd.sh) will also search java from path without a warning

As long as I am on an OS and arch having a mvnd native executable, I do not need to care that mvnd.sh is slow. mvnd.sh is fallback where some expectations are not met. But actually we could warn in mvnd.sh too. I'd prefer knowing that mvnd is missing some piece of config that makes it slower.

PR #726, opened 2022-10-21
Add property 'mvnd.preserveProjectLog' to not decorate the stdout/stderr stream

gzm55:
With the new property set to `true`, the prefix `[INFO] [stdout] ` is removed in the client terminal output. Then without `--raw-stream`, we can still get the same behavior with original maven.

### DIFF
--- a/client/src/main/java-mvnd/org/mvndaemon/mvnd/client/DefaultClient.java
+++ b/client/src/main/java-mvnd/org/mvndaemon/mvnd/client/DefaultClient.java
@@ -148,7 +148,8 @@ public class DefaultClient implements Client {
 
         int exitCode = 0;
         boolean noBuffering = batchMode || parameters.noBuffering();
-        try (TerminalOutput output = new TerminalOutput(noBuffering, parameters.rollingWindowSize(), logFile)) {
+        try (TerminalOutput output = new TerminalOutput(
+                noBuffering, parameters.rollingWindowSize(), logFile, parameters.preserveProjectLog())) {
             try {
                 final ExecutionResult result = new DefaultClient(parameters).execute(output, args);
                 exitCode = result.getExitCode();
--- a/client/src/main/java/org/mvndaemon/mvnd/client/DaemonParameters.java
+++ b/client/src/main/java/org/mvndaemon/mvnd/client/DaemonParameters.java
@@ -321,6 +321,16 @@ public class DaemonParameters {
     }
 
     /**
+     * @return <code>true</code> if the stdout/stderr project log should be preserved.
+     */
+    public boolean preserveProjectLog() {
+        return value(Environment.MVND_PRESERVE_PROJECT_LOG)
+                .orSystemProperty()
+                .orDefault()
+                .asBoolean();
+    }
+
+    /**
      * @param  newUserDir where to change the current directory to
      * @return            a new {@link DaemonParameters} with {@code userDir} set to the given {@code newUserDir}
      */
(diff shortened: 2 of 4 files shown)


gzm55:
@gnodet @ppalaga what about this pr?

ppalaga:
Could you please explain what are the possible motivations for this change? In which situations does it help? Why would user want to have this?

gzm55:
@ppalaga this option is used to keep the logging behavior the same as the original maven:

Expected from maven (no new-line char at the end):
```bash
$ mvn -DforceStdout -q help:evaluate -Dexpression=project.version -l >(sed 's/^/in log: /')
in log: 1.0.0
```

`mvnd` default:
```bash
$ mvnd -DforceStdout -q help:evaluate -Dexpression=project.version -l >(sed 's/^/in log: /')
in log: [INFO] [stdout] 1.0.0
```

`mvnd` with `--raw-streams`:
```bash
$ mvnd -DforceStdout -q help:evaluate -Dexpression=project.version -l >(sed 's/^/in log: /') --raw-streams
1.0.0
```

`mvnd` with new option `-Dmvnd.preserveProjectLog` (with new-line char at the end):
```bash
$ mvnd -DforceStdout -q help:evaluate -Dexpression=project.version -l >(sed 's/^/in log: /') -Dmvnd.preserveProjectLog
in log: 1.0.0
```

gnodet:
Note that I have the following:
```
> mvnd -q -DforceStdout help:evaluate -Dexpression=project.version --raw-streams  -l >(sed 's/^/in log: /') 2>/dev/null
1.0.0-SNAPSHOT
> mvnd -q -DforceStdout help:evaluate -Dexpression=project.version --raw-streams  1> >(sed 's/^/in log: /') 2>/dev/null
in log: 1.0.0-SNAPSHOT
```

gnodet:
> Note that I have the following:
> 
> ```
> > mvnd -q -DforceStdout help:evaluate -Dexpression=project.version --raw-streams  -l >(sed 's/^/in log: /') 2>/dev/null
> 1.0.0-SNAPSHOT
> > mvnd -q -DforceStdout help:evaluate -Dexpression=project.version --raw-streams  1> >(sed 's/^/in log: /') 2>/dev/null
> in log: 1.0.0-SNAPSHOT
> ```

So I think the problem is that mvnd implements the `-l` switch in a different way than maven.  I think fixing that will lead to a behaviour which will be closer to the maven one.

gnodet:
> > Note that I have the following:
> > ```
> > > mvnd -q -DforceStdout help:evaluate -Dexpression=project.version --raw-streams  -l >(sed 's/^/in log: /') 2>/dev/null
> > 1.0.0-SNAPSHOT
> > > mvnd -q -DforceStdout help:evaluate -Dexpression=project.version --raw-streams  1> >(sed 's/^/in log: /') 2>/dev/null
> > in log: 1.0.0-SNAPSHOT
> > ```
> 
> The problem is that mvnd implements the `-l` switch in a different way than maven. When a `-l` switch is specified, maven sets the `System.out` and `System.err` streams to the file redirection. This option is not passed to maven by mvnd, and instead, it redirects the project output to the file, but does not touch the system out and err streams. Fixing that will lead to a behaviour which will be closer to the maven one.

#779 is an attempt to fix the above problem.  It now gives the correct result (apart the added terminal newline)

gzm55:
> > > Note that I have the following:
> > > ```
> > > > mvnd -q -DforceStdout help:evaluate -Dexpression=project.version --raw-streams  -l >(sed 's/^/in log: /') 2>/dev/null
> > > 1.0.0-SNAPSHOT
> > > > mvnd -q -DforceStdout help:evaluate -Dexpression=project.version --raw-streams  1> >(sed 's/^/in log: /') 2>/dev/null
> > > in log: 1.0.0-SNAPSHOT
> > > ```
> > 
> > 
> > The problem is that mvnd implements the `-l` switch in a different way than maven. When a `-l` switch is specified, maven sets the `System.out` and `System.err` streams to the file redirection. This option is not passed to maven by mvnd, and instead, it redirects the project output to the file, but does not touch the system out and err streams. Fixing that will lead to a behaviour which will be closer to the maven one.
> 
> #779 is an attempt to fix the above problem. It now gives the correct result (apart the added terminal newline)

@gnodet the nearly same behavior of `-l` option in #779 should be better than this pr.

gnodet:
Superseded by #779

PR #729, opened 2022-10-30
Reset system properties after calling event syps

oehme:
Event spies may want to read system properties, which will fail if they are reset too early. This change makes the behavior consistent with regular Maven.


guylabs:
@ppalaga Yes exactly :) All done already. Thanks for checking back on this!

PR #734, opened 2022-11-11
Improve the README file

plamentotev:
There is wide spread misconception that in order to build a Maven project the command is `mvn clean install`.

It would be better to promote the right way (in the vast majority of the cases) to build a Maven project.

`mvnd verify` aligns with Maven documentation as well, where the example command is `mvn verify`.

p.s. Sorry for the nitpicking and thanks for the hard work. The project is amazing. Just want to fix all `mvn clean install` examples where I see them 😃

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -152,7 +152,7 @@ $ xattr -r -d com.apple.quarantine mvnd-x.y.z-darwin-amd64
 
 [source,shell]
 ----
-mvnd clean install
+mvnd verify
 ----
 
 == `mvnd` specific options


cstamas:
Stll, the `mvn clean install` works 100% always, while `mvn verify` may not, in cases like "resume from", or even if you have some nastiy plugin in picture, that expects something to be resolved from local repo (and neglects reactor), OR, when you build the plugin that is also used in same build, like here. 

So, no, there is not `s/install/verify` that is always applicable....

cesarhernandezgt:
Agree with @cstamas. I've seen cases when plugins are attached to specific phases and this gets even more interesting on multimodule projects. So finally doing a `mvn clean install` gave me the output the maintainers of the project(s) would take as `valid`.

bmarwell:
I disagree. If `mvnd verify` does not work, let's better fix the plugins instead.

I agree it's not always applicable (at this point in time), but let's promote the option which SHOULD be used most of the time.

+1 for this PR.

bmarwell (approved):

PR #736, opened 2022-11-21
Resolver 1.9.0

cstamas:
Resolver 1.9.0 has a breaking change MRESOLVER-284 that was later undone (in 1.9.1+). Hence, the
CLI changes are needd ONLY for 1.9.0 and shall be
removed on upgrade to later version.

Still, the patch is incomplete, as it merely sets
mvnd specific defaults, but does not allow
mvnd users to override locking (to HZ/redisson
for example).


cstamas:
Hm, this patch was originally made against mvnd 0.9.0 and it worked well. On master it seems both resolver get into dist (so both, resolver 1.8.2 and 1.9.0) and it causes havoc.

gnodet:
Superseded by #742

PR #764, opened 2023-01-05
Fix core export provider

oehme:
Since https://github.com/apache/maven/pull/616, the default CoreExportProvider no longer uses the provided CoreExports, but instead tries (and fails) to discover them itself.

This change fixes that by providing our own custom instance of CoreExportProvider. This allows core extension to contribute exported artifacts and exported packages again, like it used to do before the Maven 4.x upgrade.

### DIFF
--- a/integration-tests/src/test/java/org/mvndaemon/mvnd/it/ExtensionWithApiTest.java
+++ b/integration-tests/src/test/java/org/mvndaemon/mvnd/it/ExtensionWithApiTest.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.mvndaemon.mvnd.it;
+
+import javax.inject.Inject;
+
+import org.junit.jupiter.api.Test;
+import org.mvndaemon.mvnd.assertj.TestClientOutput;
+import org.mvndaemon.mvnd.client.Client;
+import org.mvndaemon.mvnd.client.DaemonParameters;
+import org.mvndaemon.mvnd.junit.MvndNativeTest;
+
+@MvndNativeTest(projectDir = "src/test/projects/extension-with-api")
+public class ExtensionWithApiTest {
+
+    @Inject
+    Client client;
+
+    @Inject
+    DaemonParameters parameters;
+
+    @Test
+    void pluginCanAccessExtensionApi() throws InterruptedException {
+        install("extension");
+        install("plugin");
+        TestClientOutput o = new TestClientOutput();
+        client.execute(o, "org.mvndaemon.mvnd.test.extension.with.api:plugin:mojo")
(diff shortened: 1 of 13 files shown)


oehme:
The underlying reason why the default logic doesn't work is because the [discovery method](https://github.com/apache/maven/blob/fc2d94f0777998000f7b886d45efb88c28fc09cd/maven-core/src/main/java/org/apache/maven/extension/internal/CoreExtensionEntry.java#L83) calls [ClassRealm.loadResourcesFromImport](https://github.com/codehaus-plexus/plexus-classworlds/blob/master/src/main/java/org/codehaus/plexus/classworlds/realm/ClassRealm.java#L526), which does not handle `META-INF` entries from imported realms. It looks up the name `META-INF/maven/extension.xml` and matches that against the import realms. The[ matching logic](https://github.com/codehaus-plexus/plexus-classworlds/blob/b17db1eae685abd72e2f2f0aad5a40c3bc09fd15/src/main/java/org/codehaus/plexus/classworlds/realm/Entry.java#L71) only accounts for classes/resources in packages, but doesn't account for META-INF. So it never never finds the extension.xml entries.

maven-embedder has the [same problem](https://github.com/apache/maven/blob/master/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java#L594) now.

ppalaga:
I have resterted the ubuntu-18.04 CI job. Not sure why it has shown up as cancelled.

gnodet:
@oehme thx for the detailed explanation.  Would it be possible to add an IT to test this feature ?

oehme:
Yes, I'll add one

oehme:
@gnodet The test infrastructure assumes that every test case is a single project, but for my case I had to install the extension first before using it in the consuming build. I hacked around this by adding the project dir as an additional parameter, but I'd love suggestions for a more elegant solution.

integration-tests/src/test/java/org/mvndaemon/mvnd/it/ExtensionWithApiTest.java
```diff
@@ -0,0 +1,64 @@
...
+        TestClientOutput o = new TestClientOutput();
+        client.execute(o, "org.mvndaemon.mvnd.test.extension.with.api:plugin:mojo")
+                .assertSuccess();
+        o.assertContainsMatchingSubsequence("Hello World!");
+    }
+
+    private void installExtension() throws InterruptedException {
+        TestClientOutput o = new TestClientOutput();
+        ((NativeTestClient) client)
+                .execute(o, parameters.multiModuleProjectDirectory().resolve("extension"), List.of("install"))
```
ppalaga, line 54: Wouldn't standard Maven `-f,--file` cli param serve the purpose?
```suggestion
                .execute(o, List.of("install", "-f", "extension"))
```

ppalaga:
> The test infrastructure assumes that every test case is a single project, but for my case I had to install the extension first before using it in the consuming build. I hacked around this by adding the project dir as an additional parameter, but I'd love suggestions for a more elegant solution.

I have left a suggestion inline.

integration-tests/src/test/java/org/mvndaemon/mvnd/it/ExtensionWithApiTest.java
oehme: Of course :) Done

oehme:
@ppalaga this should be good to go :)

ppalaga (approved):
I'll let @gnodet re-review and merge.

oehme:
Would it be possible to get an m2 for this? I'd love to run our extensions full CI build against this to make sure we work properly when 1.0 comes out.

gnodet:
> Would it be possible to get an m2 for this? I'd love to run our extensions full CI build against this to make sure we work properly when 1.0 comes out.

Yes, though I'd like to have a `maven 4.0.0-alpha-4` release and then release `mvnd 1.0.0-m2`.

ppalaga:
Here is a fix for GH Actions not publishing the PR's binaries https://github.com/apache/maven-mvnd/pull/774

gnodet (approved):

ppalaga:
@oehme, BTW if you now send some dummy PR, it will publish the distribution zips for 3 (of 4, excl. Apple M1) supported platforms  under Checks > Artifacts. You could perhaps use those to test on your side?

Issue #392, opened 2021-04-04
Define versioning policy?


rmannibucau:
Hi all,

Is it possible to get some versioning policy around mvnd.
Statements I would expect are:

1. is it using semver (if not what does mean major/minor/patch)
2. how does it follow maven and how does it relate to? ("it follows each minor of 3.x and there is a compatibility matrix for the equivalence" or "it is compatible with 3.x and rely on the env maven" kind of statement - I know both are not accurate as of today but just trying to illustrate the idea)
3. Any plan to support multiple maven branches?
4. What about security fixes in mvnd or maven stack?

Thanks.

gnodet:
> Hi all,
> 
> Is it possible to get some versioning policy around mvnd.
> Statements I would expect are:
> 
> 1. is it using semver (if not what does mean major/minor/patch)

We've been trying to follow semver, at least for minor/patch (we haven't done any major...)

> 2. how does it follow maven and how does it relate to? ("it follows each minor of 3.x and there is a compatibility matrix for the equivalence" or "it is compatible with 3.x and rely on the env maven" kind of statement - I know both are not accurate as of today but just trying to illustrate the idea)

That's definitely a problem. See below.

> 3. Any plan to support multiple maven branches?

See #384.  My original idea was to be able to support multiple maven versions managed from the same `mvnd` version.  Kinda like daemons are spawned according to the JDK, they could be spawned according to the maven version as an additional discriminating parameter (along with the JDK, the JVM options, etc...).  I do need to find some time to work on that.  It may need a few different implementations of the `DaemonMavenCli` based on which version of maven is used, but that could be one module for each maven version/branch we want to support.
This would also require some change to the distribution layout to accommodate multiple maven versions.

> 4. What about security fixes in mvnd or maven stack?

Currently, rolling up patches is quite easy.  We haven't been back porting anything at this point.  Patch releases only occur on the current master branch.  For maven, we didn't had the problem until 3.8.1, which also contains a breaking change, so it's a bit complicated.

> Thanks.

rmannibucau:
I see, is it possible to get a wiki page writing at least the semver point and that the maven relationship is "in progress" (I guess it is fine while 0.x but hope it will be clarified for >= 1.x).
If mvnd can a bit decoupled from a maven version I guess it will be simpler since it will only be about mvnd and therefore semver can be enforced so maybe let's wait for this feature to finalize this policy and close this issue?

lanmaoxinqing:
[...]

PR #769, opened 2023-01-09
Attempt at moving mvn as first class citizen in mvnd distribution, #392

gnodet:
We may need to revisit the logging configuration to be closer to maven if needed.


ppalaga (approved):

jglick:
I suspect this caused #787.

PR #784, opened 2023-01-30
Make Classworld setup more alike to vanilla Maven

oehme:
Use the plexus Launcher to start the daemon server, just like we would
launch a normal Maven process.

This improves compatbility with any extensions or plugins that assume that
their ClassLoader is a ClassRealm.


client/src/main/java/org/mvndaemon/mvnd/client/DaemonConnector.java
```diff
@@ -339,22 +339,34 @@ private Process startDaemonProcess(String daemonId, ClientOutput output) {
...
             String mvndAgentPath = null;
-            for (Path jar : jarPaths) {
-                String s = jar.getFileName().toString();
-                if (s.endsWith(".jar")) {
-                    if (s.startsWith("mvnd-common-")) {
-                        mvndCommonPath = jar.toString();
-                    } else if (s.startsWith("mvnd-agent-")) {
-                        mvndAgentPath = jar.toString();
+            String plexusClassworldsPath = null;
+            try (DirectoryStream<Path> jarPaths = Files.newDirectoryStream(mvndHome.resolve("lib/ext"))) {
```
gnodet, line 351: The implementation of `resolve` implies the usage of the file separator, which is OS specific.

common/src/main/java/org/mvndaemon/mvnd/common/MavenDaemon.java
```diff
@@ -18,67 +18,39 @@
...
-                    url = MavenDaemon.class.getClassLoader().getResource(name);
-                }
-                return url;
-            }
-        };
-        Thread.currentThread().setContextClassLoader(loader);
-        Class<?> clazz = loader.loadClass("org.mvndaemon.mvnd.daemon.Server");
-        try (AutoCloseable server = (AutoCloseable) clazz.getConstructor().newInstance()) {
+        });
+        try (InputStream in = Files.newInputStream(mvndHome.resolve("bin/m2.conf"))) {
```
gnodet, line 46: That looks wrong.  We should have a custom config that properly boots the `org.mvndaemon.mvnd.daemon.Server`.

gnodet (changes requested):

oehme:
FYI the builds are failing since https://github.com/apache/maven-mvnd/commit/bf892cd1bf10083448a9981b61615774869bf16b - Seems those artifacts aren't there yet. I'll rebase this branch on an earlier commit to get some CI feedback in the meantime.

daemon/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
```diff
@@ -473,6 +472,7 @@ DefaultPlexusContainer container() throws Exception {
 
         List<File> extClassPath = Stream.of(
                         Environment.MVND_EXT_CLASSPATH.asString().split(","))
+                .filter(s -> s != null && !s.isEmpty())
```
oehme, line 475: The missing empty filter here lead to the empty String being added to the extClassPath list, which made the classloader structure slightly different from vanilla Maven in some scenarios. Not a big deal in practice, but it broke some of our tests, so I decided to fix it for consistency :)

common/src/main/java/org/mvndaemon/mvnd/common/MavenDaemon.java
oehme: Fixed and as a result I could also remove the custom ClassRealmManager, making the code even simpler.

oehme:
That windows failure looks unrelated, it couldn't delete the daemon registry.

gnodet (approved):

lppedd:
@oehme hi! Just passing by and wondering, in which case a plugin/extension would want to explicitly check for a class loader type? Did you think about this because of some real world use case?

oehme:
Yes, the Gradle Enterprise Maven extension does some "importRealm" magic, so that other extensions (written by customers) can access its API no matter how they were registered (ext/lib, maven.ext.classpath, .mvn/extensions.xml)

PR #796, opened 2023-02-17
Provide distributions for both maven 3.9.x and 4.0.x

gnodet:


dist-m39/src/main/provisio/maven-distro.xml
```diff
@@ -0,0 +1,93 @@
...
+        <directory path="${basedir}/../dist/src/main/resources">
+            <include>platform-${os.detected.name}-${os.detected.arch}</include>
+        </directory>
+    </fileSet>
+
+    <archive name="maven-${project.version}-mvnd-mvn39-${os.detected.name}-${os.detected.arch}.zip"
+             executable="**/bin/mvnd"/>
+
+    <archive name="maven-${project.version}-mvnd-mvn39-${os.detected.name}-${os.detected.arch}.tar.gz"
+             executable="**/bin/mvnd"/>
```
psiroky, line 91: Just thinking out loud - this produces a file named e.g. `maven-1.0.0-m5-SNAPSHOT-mvnd-mvn39-linux-amd64.tar.gz` which to _me_ sounds like this is somewhat related to Maven version `1.0.0-m5-SNAPSHOT`, which is not really the case. This is the version of the Maven daemon. Was there a specific reason to change the naming scheme compared to the older branches (e.g. `maven-mvnd-0.10.0-SNAPSHOT-linux-amd64.tar.gz`)? Would it make sense to go back to `maven-mvnd-1.0.0-m5-SNAPSHOT-mvn39-linux-amd64.tar.gz`?

PR #814, opened 2023-03-13
Build on ubuntu 20.04

### LINKED ISSUE #812: Linux native binary no longer works on ubuntu 18.04 and ubuntu 20.04 (and likely any linux with glibc < 2.34)
As part of https://github.com/apache/maven-mvnd/pull/801 I upgraded the ubuntu build image from `ubuntu:18.04 to ubuntu:22.04`, mainly because GitHub is removing the support for that old ubuntu image. With that I also inadvertently bumped the minimum requirement for the Glibc the native binary needs, which now seems to be `2.34`.

Then I noticed the patch for Graal in the build script: 
```
- name: 'Patch Graal libs for only requiring glibc 2.12'
  shell: bash
  run: |
          if [[ $OS == linux ]] && [[ $GRAALVM_HOME ]] && [[ -d "$GRAALVM_HOME/lib/static/linux-amd64/glibc" ]]; then
            mkdir -p client/target/graalvm-libs-for-glibc-2.12
            echo "memcpy memcpy@GLIBC_2.2.5" >client/target/glibc.redef
            echo "posix_spawn posix_spawn@GLIBC_2.2.5" >>client/target/glibc.redef
            find "$GRAALVM_HOME/lib/static/linux-amd64/glibc" -name '*.a' | while IFS= read -r input; do
              output="client/target/graalvm-libs-for-glibc-2.12/$(basename -- "$input")"
              objcopy --redefine-syms=client/target/glibc.redef -- "$input" "$output" 2>/dev/null
            done
            find /usr/lib -name libz.a | xargs -r -I {} objcopy --redefine-syms=client/target/glibc.redef {} client/target/graalvm-libs-for-glibc-2.12/libz.a
          fi
```
The patch suggests that `mvnd` should support glibc 2.12, which is no longer the case.

This the output I get when running on ubuntu 18.04 and ubuntu 20.04, with latest binaries from master:
```
/tmp/mvnd-test ❯ podman run -it -v .:/mvnd ubuntu:18.04 bash                                                                                                                                                                                                                                                          
root@542c2dc5feba:/# ./mvnd/maven-1.0.0-m5-SNAPSHOT-mvnd-m39-linux-amd64/bin/mvnd --version
./mvnd/maven-1.0.0-m5-SNAPSHOT-mvnd-m39-linux-amd64/bin/mvnd: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./mvnd/maven-1.0.0-m5-SNAPSHOT-mvnd-m39-linux-amd64/bin/mvnd)
./mvnd/maven-1.0.0-m5-SNAPSHOT-mvnd-m39-linux-amd64/bin/mvnd: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./mvnd/maven-1.0.0-m5-SNAPSHOT-mvnd-m39-linux-amd64/bin/mvnd)
```

Now, I am not sure about the intended supportability for different glibc versions. Ubuntu 18.04 is already pretty old and according to https://wiki.ubuntu.com/Releases, the end of standard support is like a month away. So maybe that one could be dropped? But then it seems like `mvnd` should still support at least ubuntu 20.04, which will be supported for another two years.

psiroky:
* fixes #812 
* building on ubuntu:22.04 produces binaries which require too recent glibc
   version (2.32+) and that means the binary no longer works on
   older ubuntu versions, like 20.04 and 18.04.

 * building on ubuntu:20.04 fixes the problem and the binary again works
   on ubuntu 18.04 and 20.04

 * also bumped the version in the release workflow which got forgotten

### DIFF
--- a/.github/workflows/early-access.yaml
+++ b/.github/workflows/early-access.yaml
@@ -31,7 +31,7 @@ jobs:
   default-build:
     name: 'Default build (without Graal)'
     if: startsWith(github.event.head_commit.message, '[release] Release ') != true
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
     steps:
       - name: 'Checkout'
         uses: actions/checkout@v2
@@ -52,7 +52,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, macOS-10.15, windows-2019 ]
+        os: [ ubuntu-20.04, macOS-10.15, windows-2019 ]
     runs-on: ${{ matrix.os }}
 
     steps:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ ubuntu-18.04, macOS-10.15, windows-2019 ]
+        os: [ ubuntu-20.04, macOS-10.15, windows-2019 ]
     runs-on: ${{ matrix.os }}
 
     steps:
@@ -88,7 +88,7 @@ jobs:
 
   source:
     name: 'Build source distributions'
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
       - name: 'Check out repository'
         uses: actions/checkout@v2
@@ -121,7 +121,7 @@ jobs:
             target/maven-mvnd-*.tar.gz
 
   release:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     needs: [build, source]
 
     steps:


psiroky:
Hmm, so this moves the glibc version requirement to 2.17, which is better than before, but still not quite the same

```
./mvnd/maven-1.0.0-m5-SNAPSHOT-mvnd-m39-linux-amd64/bin/mvnd: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.17' not found (required by ./mvnd/maven-1.0.0-m5-SNAPSHOT-mvnd-m39-linux-amd64/bin/mvnd)
```

psiroky:
Reading through https://github.com/apache/maven-mvnd/pull/730, it says
>An explicit glibc requirement also removes a side effect of changing the build system. When update the build system from ubuntu 18.04 to 20.04, the glibc requirement may also increase to match the new os. On the other hand, glibc 2.12 is required by graalvm.

which is not really the case (anymore?). We bumped the build system and as a side-effect glibc version got bumped as well. GraalVM got updated from 22.2.0 to 22.3.1 in the meantime, in case that could be somehow related.

psiroky:
I wasn't able to make this work so that the glibc requirement would still be 2.12 (released on 2010-12-13). I tried to follow the similar approach with `glibc.redef` as there already is, but that feels like magic, redefining symbols is "random" libraries. I believe this is also going to break again when we upgrade the build system.

There is a library which is using `GLIBC_2.17  clock_gettime` (not even sure which one to be honest) and I am not sure how to get rid of that.

I double-checked and CentOS 6 is indeed using glibc 2.12, but it is also end-of-life for more than two years now (https://wiki.centos.org/HowTos/EOL). Maybe it would make sense to bump the glibc version anyway and support 2.17, which is the one used in CentOS 7?

PR #834, opened 2023-04-06
Fix PluginDescriptorCache

oehme:
The second get-method wasn't actually overloaded, but had a different signature than the superclass. This resulted in the superclass cache being filled and never cleared, even when flush is called.


oehme:
I wonder why we're extending the default class at all? Are there direct references to the default class in Maven core?

gnodet:
> I wonder why we're extending the default class at all? Are there direct references to the default class in Maven core?

To avoid duplicating the whole class, including the `CacheKey` inner class ?

oehme:
Ah yes, that makes sense.

PR #866, opened 2023-07-03
Fix interpolation of environment variables in config properties (fixes #693)

### LINKED ISSUE #693: can't use system environment variable?
execute `mvnd clean` `mvnd -version`,tip me 'Could not get a real path from path $JAVA_HOME'
image

gnodet:


gnodet:
Supersedes #857

Issue #870, opened 2023-07-25
Maven 4 flavour does not use full terminal


mthmulders:
One of the many things I like in Maven 4 (alpha) is the longer line width. As an example, see the first few lines of any build:

```
Apache Maven 4.0.0-alpha-8-SNAPSHOT (36db1e35cac5e8c72bf0c795dd08756e50079e05)
Maven home: /opt/homebrew/Cellar/maven-snapshot/4.0.0-alpha-8-20230719.153755-14/libexec
Java version: 17.0.7, vendor: Eclipse Adoptium, runtime: /Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home
Default locale: en_NL, platform encoding: UTF-8
OS name: "mac os x", version: "13.4.1", arch: "aarch64", family: "mac"
[INFO] Scanning for projects...
[INFO] --------------------------------------------------------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO]
```

The equivalent with mvnd however does not leverage the full terminal width:

```
Apache Maven Daemon (mvnd) 1.0-m7 darwin-aarch64 native client (b2ef5d81997adbcdb72dc8c5603722538fa641fe)
Terminal: org.jline.terminal.impl.PosixSysTerminal with pty org.jline.terminal.impl.jansi.osx.OsXNativePty
Apache Maven 4.0.0-alpha-7 (bf699a388cc04b8e4088226ba09a403b68de6b7b)
Maven home: /Users/maarten/Tools/maven-mvnd-1.0-m7-m40-darwin-aarch64/mvn                                                                                                                                          Java version: 17.0.7, vendor: Eclipse Adoptium, runtime: /Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home
Default locale: en_NL, platform encoding: UTF-8
OS name: "mac os x", version: "13.4.1", arch: "aarch64", family: "mac"
[INFO] Processing build on daemon 93e78205
[INFO] Scanning for projects...
[INFO] BuildTimeEventSpy is registered.
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO]
```

In both cases, I ran the `validate` phase and added `-V` to print the version of Maven (Daemon).

gnodet:
[...]

PR #889, opened 2023-10-14
Use full terminal width in CLI with Maven 4

KasNotten:
I am trying to solve the following issue: #870 

As far as I can determine this code (Also contains some test code which can be removed later) should work for using the full terminal width in maven-mvnd, comparing it with the code in the maven repository for using the full terminal width in maven 4. 

However, the terminal width is always -1.

Can anyone shed some light on as to why this is happening? Maybe @ppalaga or @gnodet ?


gnodet:
Your PR is wrong.  Jansi cannot be used as there's no terminal.  I'll try to investigate.

gnodet:
The terminal width is sent by the client.
A proposed fix is available at https://github.com/apache/maven-mvnd/pull/891

PR #890, opened 2023-10-14
Update build instructions for Windows

KasNotten:
I would like to propose adding the following details to the instructions for building `maven-mvnd` on Windows.
I had to spend some time figuring out which tools I needed and I figured I would save other future Windows contributors the trouble.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -227,6 +227,9 @@ $ ls -lh target/mvnd
 -rwxrwxr-x. 1 ppalaga ppalaga 25M Jun  2 13:23 target/mvnd
 ----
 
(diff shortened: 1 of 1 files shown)


mthmulders:
It confuses me that the build fails, but only on Ubuntu, for something that's only a documentation change. @gnodet / @ppalaga, any objections on merging it, despite the failing build?

gnodet:
> It confuses me that the build fails, but only on Ubuntu, for something that's only a documentation change. @gnodet / @ppalaga, any objections on merging it, despite the failing build?

Some tests are flaky.  When that's the case, I usually re-run them to have green lights...

gnodet (approved):

PR #925, opened 2024-03-14
Update README.adoc

nathansit:
* Just fixed some typos
* Adjusted some wording
* Feel free to accept whichever parts make sense (if any)

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -35,8 +35,8 @@ This architecture brings the following advantages:
 
 * By default, `mvnd` is building your modules in parallel using multiple CPU cores. The number of utilized cores is
   given by the formula `Math.max(Runtime.getRuntime().availableProcessors() - 1, 1)`. If your source tree does not
-  support parallel builds, pass `-T1` on the command line to make your build serial.
-* Improved console output: we believe that the output of a parallel build on a stock Maven is hard to follow. Therefore,
+  support parallel builds, pass `-T1` into the command line to make your build serial.
+* Improved console output: we believe that the output of a parallel build on stock Maven is hard to follow. Therefore,
 we implemented a simplified non-rolling view showing the status of each build thread on a separate line. This is
 what it looks like on a machine with 24 cores:
 +
@@ -70,7 +70,7 @@ $ brew install mvndaemon/homebrew-mvnd/mvnd
 
 We're looking for contribution to support https://www.macports.org[MacPorts],
 https://community.chocolatey.org/packages/mvndaemon/[Chocolatey], https://scoop.sh/[Scoop] or
-https://github.com/joschi/asdf-mvnd#install[asdf].  If you're fancy helping us...
+https://github.com/joschi/asdf-mvnd#install[asdf].  If you fancy helping us...
 
 ////
 === Install using https://www.macports.org[MacPorts]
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -24,8 +24,7 @@ This architecture brings the following advantages:
 * The JVM for running the actual builds does not need to get started anew for each build.
 * The classloaders holding classes of Maven plugins are cached over multiple builds. The plugin jars are thus read
   and parsed just once. SNAPSHOT versions of Maven plugins are not cached.
-* The native code produced by the Just-In-Time (JIT) compiler inside the JVM is kept too. Compared to stock Maven,
-  less time is spent by the JIT compilation. During the repeated builds the JIT-optimized code is available
+* The native code produced by the Just-In-Time (JIT) compiler inside the JVM is kept too. JIT compilation is faster than stock Maven. During the repeated builds the JIT-optimized code is available
```
ppalaga, line 27: > JIT compilation is faster than stock Maven.

This changes the meaning in a wrong way. The changed sentence does not state anything useful.

README.adoc
```diff
@@ -70,7 +69,7 @@ $ brew install mvndaemon/homebrew-mvnd/mvnd
 
 We're looking for contribution to support https://www.macports.org[MacPorts],
 https://community.chocolatey.org/packages/mvndaemon/[Chocolatey], https://scoop.sh/[Scoop] or
-https://github.com/joschi/asdf-mvnd#install[asdf].  If you're fancy helping us...
+https://github.com/joschi/asdf-mvnd#install[asdf].  If you fancy helping us...
```
ppalaga, line 72: -1 on this change. The original was correct IMO.

ppalaga (changes requested):

README.adoc
Stephan202: I'm not a native speaker, but IIUC the original is incorrect, and the suggestion _is_ correct. See also examples e.g. [here](https://dictionary.cambridge.org/dictionary/english/fancy).

nathansit: I can see it being written as "If you are fancy, help us" or something along those lines. But I figured that the correction was your intended turn of phrase.

[Another example](https://hinative.com/questions/16176373) here the answers seem to change the you're to you

ppalaga: Thanks for pointing to an evidence, gentlemen. I did not know "fancy" was also a verb.

README.adoc
```diff
@@ -183,7 +183,7 @@ Properties defined in the first files will take precedence over properties speci
 
 A few special properties do not follow the above mechanism:
 
-* `mvnd.daemonStorage`: this property defines the location where mvnd stores its files (registry and daemon logs).  This property can only be defined as a system property on the command line
+* `mvnd.daemonStorage`: this property defines the location where mvnd stores its files (registry and daemon logs).  This property can only be defined as a system property into the command line
```
Stephan202, line 186: I think that "on" is more idiomatic in this context :upside_down_face: (Not sure I've ever seen "into" used this way.)

Stephan202 (commented):
The other two changes LGTM :)

README.adoc
nathansit: Aha good catch I must have overwritten and then modified the wrong "on" when I force pushed.

ppalaga (approved):
Thanks  @nathansit!

Issue #929, opened 2024-04-12
TerminalOutput.pathToMaven throws StringIndexOutOfBoundsException


ppalaga:
Steps to reproduce:

```
git clone git@github.com:quarkiverse/quarkus-langchain4j.git
cd quarkus-langchain4j
git reset --hard 177efc97b49823ad747a87766ea1e1d0c2b71ef6

mvnd --stop
mvnd -v 
Apache Maven Daemon (mvnd) 1.0-m8 linux-amd64 native client (0f4bdb6df5e74453d8d558d292789da4e66a7933)Terminal: org.jline.terminal.impl.PosixSysTerminal with pty org.jline.terminal.impl.jansi.linux.LinuxNativePty
Apache Maven 3.9.5 (57804ffe001d7215b5e7bcb531cf83df38f93546)
Maven home: /home/ppalaga/.sdkman/candidates/mvnd/1.0-m8-m39/mvn
Java version: 17.0.9, vendor: Eclipse Adoptium, runtime: /home/ppalaga/.sdkman/candidates/java/17.0.9-tem
Default locale: en_IE, platform encoding: UTF-8
OS name: "linux", version: "6.4.11-200.fc38.x86_64", arch: "amd64", family: "unix"

rm -Rf /tmp/empty-local-maven-repo
mvnd clean install -DskipTests -Dquarkus.build.skip -Dmaven.repo.local=/tmp/empty-local-maven-repo
...
Exception in thread "main" java.lang.StringIndexOutOfBoundsException: String index out of range: 17
        at java.base@17.0.6/java.lang.StringLatin1.charAt(StringLatin1.java:48)
        at java.base@17.0.6/java.lang.String.charAt(String.java:1513)
        at org.mvndaemon.mvnd.common.logging.TerminalOutput.pathToMaven(TerminalOutput.java:724)
        at org.mvndaemon.mvnd.common.logging.TerminalOutput.formatTransfers(TerminalOutput.java:697)
        at org.mvndaemon.mvnd.common.logging.TerminalOutput.addProjectLine(TerminalOutput.java:810)
        at org.mvndaemon.mvnd.common.logging.TerminalOutput.update(TerminalOutput.java:615)
        at org.mvndaemon.mvnd.common.logging.TerminalOutput.accept(TerminalOutput.java:217)
        at org.mvndaemon.mvnd.client.DefaultClient.execute(DefaultClient.java:366)
[...]

PR #930, opened 2024-04-12
Make TerminalOutput.pathToMaven() more robust to avoid #929

ppalaga:
Mitigates #929 but does not have an ambition to find the root cause.

### DIFF
--- a/common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
@@ -714,37 +714,45 @@ public class TerminalOutput implements ClientOutput {
 
     public static String pathToMaven(String location) {
         String[] p = location.split("/");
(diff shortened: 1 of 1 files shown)


gastaldi:
It would be nice to log (debug?) when the URL is malformed

ppalaga:
> It would be nice to log (debug?) when the URL is malformed

Well, the issue we are mitigating happens only in the text UI, where the place is sparse and where there is no concept of log level. I am not sure the information that we were not able to parse some URL is that interesting to end users. If we cannot parse it, we just print it as is. If the transfer failure matters, the daemon will inform them through a separate message. If it is ignorable, I do not think they need to know. My 2c.

gnodet (approved):
LGTM.  Next maven resolver will probably have a better way to access this information with https://github.com/apache/maven-resolver/pull/474, so we'll be able to update to that one when it's integrated into 3.9.x and 4.x maven branches.

PR #950, opened 2024-04-18
Cumulative update of GH flow

cstamas:
Instead to do this one-by-one as pushed by dependendabot...

### DIFF
--- a/.github/workflows/early-access.yaml
+++ b/.github/workflows/early-access.yaml
@@ -18,13 +18,9 @@
 name: Early Access
 
 # trigger on push to branches and PR
-on:
-  push:
-    branch:
-  pull_request:
+on: [push, pull_request]
 
 env:
-  GRAALVM_VERSION: '22.3.1'
   JAVA_VERSION: '17'
 
 jobs:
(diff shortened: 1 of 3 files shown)


cstamas:
related
https://github.com/graalvm/setup-graalvm?tab=readme-ov-file#migrating-from-graalvm-223-or-earlier-to-the-new-graalvm-for-jdk-17-and-later

.github/workflows/early-access.yaml
```diff
@@ -52,12 +57,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, macOS-11, windows-2019 ]
+        os: [ ubuntu, macos, windows ]
```
slawekjaranowski, line 60: There are not runners like `ubuntu`
There are `ubuntu-latest` and so on ...

https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners

cstamas:
@gzm55 help! :smile:

gzm55:
> @gzm55 help! 😄

@cstamas try append this line to client/src/main/resources/glibc/glibc.redef

```
pthread_mutex_trylock pthread_mutex_trylock@GLIBC_2.2.5
```

gzm55:
> @gzm55 Thanks a lot! It seems it worked, ubuntu job is green...

yeah, new graalvm introduce new api dependency.

gnodet (approved):

PR #952, opened 2024-04-19
Upgrade develocity

cstamas:


gnodet:
So the XML config needs to be updated as well ?

gnodet:
@ppalaga @cstamas given the enterprise gradle extension do not work with Maven 4, I'm tempted to just remove those. wdyt ?

cstamas:
Agreed

PR #953, opened 2024-04-20
Default value must be manually handled

### LINKED ISSUE #912: Detect smart builder in .mvn/extensions.xml and ignore it
As currently it fails:
```
class io.takari.maven.builder.smart.DependencyGraph cannot be cast to class io.takari.maven.builder.smart.DependencyGraph (io.takari.maven.builder.smart.DependencyGraph is in unnamed module of loader org.codehaus.plexus.classworlds.realm.ClassRealm @97e1986; io.takari.maven.builder.smart.DependencyGraph is in unnamed module of loader org.codehaus.plexus.classworlds.realm.ClassRealm @63bde6c2)
```

cstamas:
Default must be manually handled, i missed this.

Fixes #912


wendigo:
That works :) Thanks!

cstamas:
Am adding IT (something I'd done initially would catch this), and will merge

wendigo (approved):

CrazyHZM (approved):

cstamas:
The code change is OK and fixes issue, but something is wrong.
Created https://github.com/apache/maven-mvnd/issues/957

wendigo:
Can we get this in for alpha14?

cstamas:
Sure, this will be in it. But as i wrote, seems there is something else to be fixed as IT this PR adds shows...

cstamas:
@gnodet If you run this PR locally:
* it will pass ok, but
* maven-conf-ignore-ext IT is **wrong**: maven.conf where ignore is, is not observed by mvnd. Also local repository will show that extension foo:bar is attempted to resolve, but naturally unsuccessfully
* this means that mvnd happily goes along instead to fail (as it failed to resolve the extension)

Any idea?

client/src/main/java/org/mvndaemon/mvnd/client/DaemonParameters.java
```diff
@@ -491,16 +491,22 @@ private static List<CoreExtension> readCoreExtensionsDescriptor(Path multiModule
 
     private static List<CoreExtension> filterCoreExtensions(List<CoreExtension> coreExtensions) {
         Set<String> exclusions = new HashSet<>();
-        String exclusionsString =
-                systemProperty(Environment.MVND_CORE_EXTENSIONS_EXCLUDE).asString();
+        String exclusionsString = systemProperty(Environment.MVND_CORE_EXTENSIONS_EXCLUDE)
+                .asOptional()
```
gnodet, line 495: I think this should be:
```
String exclusionsString =
                systemProperty(Environment.MVND_CORE_EXTENSIONS_EXCLUDE).orDefault().asString();
```

client/src/main/java/org/mvndaemon/mvnd/client/DaemonParameters.java
```diff
@@ -491,16 +491,22 @@ private static List<CoreExtension> readCoreExtensionsDescriptor(Path multiModule
...
         if (exclusionsString != null) {
             exclusions.addAll(Arrays.stream(exclusionsString.split(","))
                     .filter(e -> e != null && !e.trim().isEmpty())
                     .collect(Collectors.toList()));
         }
-        return coreExtensions.stream()
-                .filter(e -> !exclusions.contains(e.getGroupId() + ":" + e.getArtifactId()))
-                .collect(Collectors.toList());
+        if (!exclusions.isEmpty()) {
+            LOG.info("Excluded extensions (GA): {}", exclusions);
```
gnodet, line 503: This should definitely not be part of the output, and not 4 times.
I'll revert this chunk, or log to debug at most.

integration-tests/src/test/projects/maven-conf-ignore-ext-def/pom.xml
```diff
@@ -0,0 +1,27 @@
+<!--
+
+    Copyright 2019-2022 the original author or authors.
```
gnodet, line 3: Remove the copyright ?

integration-tests/src/test/projects/maven-conf-ignore-ext/pom.xml
```diff
@@ -0,0 +1,27 @@
+<!--
+
+    Copyright 2019-2022 the original author or authors.
```
gnodet, line 3: Maybe remove the copyright ?

gnodet (changes requested):

integration-tests/src/test/projects/maven-conf-ignore-ext-def/pom.xml
cstamas: every IT POM has it.... so maybe a sweeping change to remove them all and leave plain ASL2 header?

cstamas:
Unsure why ubuntu failed, win passed.... locally also passed w/ same CLI params as CI invoked :thinking:

common/src/main/java/org/mvndaemon/mvnd/common/Environment.java
```diff
@@ -224,7 +224,7 @@ public enum Environment {
             null,
             "io.takari.maven:takari-smart-builder",
             OptionType.STRING,
-            Flags.OPTIONAL),
+            Flags.DISCRIMINATING | Flags.OPTIONAL),
```
gnodet, line 227: I don't think this one is discriminating.  What is discriminating is the list of extensions to load which may actually be affected by this property, but not necessarily.  If you add an exclusion which is not supposed to be loaded, this won't affect the result.  In addition, that property is not really interesting on the daemon side, as the extensions list is loaded by the client...

cstamas: removed

PR #956, opened 2024-04-20
Make our lives simpler: mvnd is Java17

### LINKED ISSUE #954: Make mvnd Java17
Fully. So both m39 and m40 should require Java17.
This is basically to simplify our own lives, and easier maintain. Maven4 is already Java17. 
If someone must have Java8 to build, should use Maven 3.9.x + smart builder.

cstamas:
To ease our lives, mvnd is Java17. m40 is already Java17, and now m39 requires it as well.

This is okay constraint, as it someone must have Java8 running Maven 3.9.x, Maven 3.9.x + smart builder is still there and works on Java8.

Fixes #954


gnodet:
> To ease our lives, mvnd is Java17. m40 is already Java17, and now m39 requires it as well.
> 
> This is okay constraint, as it someone must have Java8 running Maven 3.9.x, Maven 3.9.x + smart builder is still there and works on Java8.
> 
> Fixes #954

I disagree, it's there and it's working. I don't think anyone had any problem with it since its introduction.   If we do split again between 3.9.x and 4.x support, then we can surely drop some stuff from the 4.x branch.
Also, this PR seems to remove the java fallback which is used to make mvnd work on platforms which are not supported by the native builds.

cstamas:
But, locally am not building native, so how does mvnd work for me then?

gnodet:
> But, locally am not building native, so how does mvnd work for me then?

The native binary is preferred when available.  Then `java-mvnd` is used to run mvnd in JVM mode and JDK >= 11.  The `java-fallback` is used in JVM mode when JDK < 11 and basically falls back to plain maven.  The goal is to be able to always use mvnd instead of maven.  You don't have to care, it will just work whatever your environment is.  What your PR propose, is, of course, a simplification of the code, but it will limit the usage of mvnd to a smaller set of supported combinations of platforms + JDK.

cstamas:
Agreed, closing this one out

PR #969, opened 2024-04-23
Daemons should not swallow any exception

### LINKED ISSUE #957: mvnd happily fails to load extension and continues on
As PR shows:
https://github.com/apache/maven-mvnd/pull/953

### LINKED ISSUE #966: Mvnd should fail is extension cannot be loaded
> @gnodet If you run this PR locally:
> 
> * it will pass ok, but
> * maven-conf-ignore-ext IT is **wrong**: maven.conf where ignore is, is not observed by mvnd. Also local repository will show that extension foo:bar is attempted to resolve, but naturally unsuccessfully
> * this means that mvnd happily goes along instead to fail (as it failed to resolve the extension)
> 
> Any idea?

Yes, the problem is here:
* https://github.com/apache/maven-mvnd/blob/dc4179fc3ba193e18d84dc090abfe336649d63ed/daemon-m40/src/main/java/org/apache/maven/cli/DaemonMavenCli.java#L699
* https://github.com/apache/maven-mvnd/blob/dc4179fc3ba193e18d84dc090abfe336649d63ed/daemon-m39/src/main/java/org/apache/maven/cli/DaemonMavenCli.java#L615
A failure to resolve the extension simply prints a warning and ignore.... That should be fixed.

_Originally posted by @gnodet in https://github.com/apache/maven-mvnd/issues/953#issuecomment-2070825470_

### LINKED ISSUE #968: maven-git-versioning-extension extension is not supported
maven-git-versioning-extension
```xml
<extension>
    <groupId>me.qoomon</groupId>
    <artifactId>maven-git-versioning-extension</artifactId>
    <version>9.7.0</version>
</extension>
```

I want to disable maven-git-version-extension with the `-Dversion.disable=true` configuration, but when using mvnd, 
the `-Dversion.disable=true` configuration does not work. But with the mvn command, this configuration works.

Does mvnd not yet support Maven-git-version-extension???

cstamas:
The "happy path" should happen without anything thrown, unsure why here runtime ex is rethrown, others swallowed.

Change: just remove the try/catch and make method throw Exception (diff best seen with whitespace ignored).

Fixes #957
Fixes #966

### DIFF
--- a/daemon-m39/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
+++ b/daemon-m39/src/main/java/org/apache/maven/cli/DaemonMavenCli.java
@@ -717,64 +717,56 @@ public class DaemonMavenCli implements DaemonCli {
     }
 
     private List<CoreExtensionEntry> loadCoreExtensions(
-            List<CoreExtension> extensions, ClassRealm containerRealm, Set<String> providedArtifacts) {
-        try {
-            if (extensions.isEmpty()) {
-                return Collections.emptyList();
-            }
-            ContainerConfiguration cc = new DefaultContainerConfiguration() //
-                    .setClassWorld(classWorld) //
-                    .setRealm(containerRealm) //
-                    .setClassPathScanning(PlexusConstants.SCANNING_INDEX) //
-                    .setAutoWiring(true) //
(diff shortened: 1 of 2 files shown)


gnodet (approved):
LGTM

PR #1017, opened 2024-06-13
Fix deprecation warning about JAnsi Terminal

cstamas:
And use JNI (and FFM) instead.

Changes:
* use DefaultClient from master (thanks @gnodet !)
* drop jansi (as DefaultClient should not depend on it)
* introduce two new jline3 terminal backends: jni and ffm (used on Java 22+)
* tested/verified (thanks @wendigo )


wendigo (approved):
LGTM!

gnodet (approved):

PR #1020, opened 2024-06-14
More script changes

cstamas:
Changes:
* `release-build.sh`: Just do not push. This allows one to locally verify the outcome, and push (for now manually like `git push origin mvnd-1.x --tags`). Added reminder message.
* `release.yaml`: whenever mvn would resolve deps, make sure release settings XML is present. It does not stir anything if Maven is in Central, but as in this moment, maven is still staged only.

### DIFF
--- a/build/release-build.sh
+++ b/build/release-build.sh
@@ -113,7 +113,7 @@ startup_check
 # update version
 mvn versions:set -DnewVersion=$VERSION
 
-# udpate changelog
+# update changelog
 docker run -it --rm -v "$(pwd)":/usr/local/src/your-app githubchangeloggenerator/github-changelog-generator \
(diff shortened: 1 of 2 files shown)


gnodet (approved):

build/release-build.sh
```diff
@@ -139,4 +139,6 @@ mvn versions:set -DnewVersion=$NEXT_VERSION
 # commit
 git add -A
 git commit -m "Next is $NEXT_VERSION"
-git push origin mvnd-1.x
+# git push origin mvnd-1.x
+
+# NOTE: Manually push once done: git push origin mvnd-1.x --tags
```
gnodet, line 144: Can we echo that somehow ?

cstamas: added reminder message at end

PR #1041, opened 2024-06-20
Lower CPU requirements

### LINKED ISSUE #1029: Cannot use mvnd due to missing CPU features (on an older machine)
Thank you very much for mvnd! I'm running the prebuilt binary just fine on my AMD Zen2 laptop, but when I tried the same on an older machine, I only got a sad message about a target architecture mismatch:
```
The current machine does not support all of the following CPU features that are required by the image: [CX8, CMOV, FXSR, MMX, SSE, SSE2, SSE3, SSSE3, SSE4_1, SSE4_2, POPCNT, LZCNT, AVX, AVX2, BMI1, BMI2, FMA].
Please rebuild the executable with an appropriate setting of the -march option.
```
In my case the CPU (a SandyBridge box that still works perfectly fine) is missing LZCNT, AVX2, BMI1/2, FMA.

I know this is an old system, but it still works fine and in order to reach the widest audience it would IMHO be great if future releases could be built with a more conservative `-march` target.
mvnd hardly benefits from BMI/FMA and probably only in a few synthetic benchmarks from AVX2, so this would not have any measurable performance impact.

cstamas:
Using "compatibility" march.

Fixes #1029

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -228,6 +228,7 @@
               <mainClass>org.mvndaemon.mvnd.client.DefaultClient</mainClass>
               <imageName>mvnd</imageName>
               <buildArgs>--no-fallback
+                         -march=compatibility
                          -H:+UnlockExperimentalVMOptions
                          ${graalvm-native-static-opt}
                          ${graalvm-native-glibc-opt}


hhoffstaette:
I tested the binary from the CI run and it works. :raised_hands:
x86-64-v2 would have been fine too but whatever; thank you very much! :heart:

PR #1046, opened 2024-06-21
Move exception to a newline after a message

welandaz:
This PR adds a line separator, which moves the exception to a new line instead of having it be concatenated.

Before the change, an example error would've been logged like this:
`[main] INFO test - info with exceptionjava.lang.RuntimeException: test exception`

After the change, the exception is separated with a new line:
```
[main] INFO test - info with exception
java.lang.RuntimeException: test exception
```

There is no test for this class, and I didn't find any other unit tests (only the integration ones), so I didn't create one.
I can set it up if needed, although a testing framework would have to be introduced too, as it's missing now afaik.


gnodet:
Duplicated by #1058

PR #1071, opened 2024-07-16
Fix daemon connection issues

oehme:
We have a test suite for a Maven extension that runs around 5000 builds and before this change, on each CI build, about 100 of them would fail with daemon connection issues. The symptom would always be the same - The daemon is waiting for the build request, the client has already sent it, but the daemon never receives it, no matter how long we configured the timeouts. 

The reason turned out to be the way that the daemon waits for the request:

The Server was using a SynchrnousQueue to coordinate the main thread
and the background thread that receives the request from the client.
A SynchronousQueue only allows insertions when a corresponding call
to `get` is in progress. However, since the receiver thread is started
before the call to `get`, there was a short time window, where the call
to `queue.offer` could fail and simply return `false`. This return code
was ignored.

A possible solution would have been to call `put` instead of `offer`,
but I decided to replace the queue with a Future, since we only wait
for a single element.

### DIFF
--- a/daemon/src/main/java/org/mvndaemon/mvnd/daemon/Server.java
+++ b/daemon/src/main/java/org/mvndaemon/mvnd/daemon/Server.java
@@ -37,11 +37,11 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.PriorityBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Condition;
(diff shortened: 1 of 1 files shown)


gnodet:
It seems much less stable on Windows for some reason...

oehme:
Yep, looks like I'll have to dig into the Windows behavior some more. Taking back into draft.

oehme:
After a lot of debugging, I've finally figured out the true cause of this, see the updated description.

I guess my previous attempt had simply changed the timing slightly, reducing the flakiness on my machine, but increasing it in the Windows builds, which explains those failures.

gnodet:
> After a lot of debugging, I've finally figured out the true cause of this, see the updated description.
> 
> I guess my previous attempt had simply changed the timing slightly, reducing the flakiness on my machine, but increasing it in the Windows builds, which explains those failures.

Really nice catch.

However, even if guava is a transitive dependencies, it's not really a direct dependency in maven.
I would really prefer if we could use a JDK collection somehow, maybe `new ArrayBlockingQueue(1)` ?

oehme:
Good point, changed to CompletableFuture

ppalaga:
Indeed, great catch @oehme 🙏

PR #1097, opened 2024-08-14
Fix SDKMAN publish from master

cstamas:
mvnd on master still produces "betas" (as long as mvn master does), and it must not be set as "default" version in SDKMAN. Defaults are still coming from mvnd-1.x branch.


gnodet (approved):
No problem with the principle, we kinda knew it could be seen as a a problem.

PR #1104, opened 2024-08-19
chore: make install commands copy+pastable

kevdowney:
Readme cleanup to allow the install commands to be copied and pasted without modification.

This fixes the error:
```
zsh: command not found: $
```


cstamas (approved):

alexec:
👋 @kevdowney

slachiewicz (approved):

PR #1175, opened 2024-10-18
Switch the build to using Maven 4 and the 4.1.0 model

gnodet:
This PR obviously needs a CI change to build with 4.0.0-beta-5, but it serves as a testing for https://github.com/apache/maven/pull/1818

### DIFF
--- a/agent/pom.xml
+++ b/agent/pom.xml
@@ -16,14 +16,9 @@
     limitations under the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.1.0 https://maven.apache.org/xsd/maven-4.1.0.xsd">
 
-  <modelVersion>4.0.0</modelVersion>
-  <parent>
-    <groupId>org.apache.maven.daemon</groupId>
-    <artifactId>mvnd</artifactId>
-    <version>2.0.0-beta-2-SNAPSHOT</version>
-  </parent>
+  <parent />
 
   <artifactId>mvnd-agent</artifactId>
 
--- a/build-plugin/pom.xml
+++ b/build-plugin/pom.xml
@@ -16,14 +16,9 @@
     limitations under the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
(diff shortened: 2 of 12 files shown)


cstamas:
BOM GA's changed, this PR will now fail with beta-5 unless updated

agent/pom.xml
```diff
@@ -16,14 +16,10 @@
     limitations under the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.1.0 http://maven.apache.org/maven-v4_1_0.xsd">
```
mthmulders, line 19: I think the correct location for the XSD is https://maven.apache.org/xsd/maven-4.1.0-beta-4.xsd. The current URL gives a HTML page with "Not found".

agent/pom.xml
```diff
@@ -16,14 +16,10 @@
...
 
-  <modelVersion>4.0.0</modelVersion>
-  <parent>
-    <groupId>org.apache.maven.daemon</groupId>
-    <artifactId>mvnd</artifactId>
-    <version>2.0.0-beta-2-SNAPSHOT</version>
-  </parent>
+  <modelVersion>4.1.0</modelVersion>
```
mthmulders, line 21: This can be inferred from the XSD, so we could drop it.

build-plugin/pom.xml
mthmulders, line 1: Same as **agent/pom.xml**.

client/pom.xml
mthmulders, line 1: Same as **agent/pom.xml**.

common/pom.xml
```diff
@@ -55,11 +51,6 @@
       <artifactId>junit-jupiter</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
```
mthmulders, line 58: Is this dependency not necessary anymore? The change seems unrelated to me...

common/pom.xml
mthmulders, line 1: Same as **agent/pom.xml**.

pom.xml
```diff
@@ -239,51 +134,15 @@
...
-        <artifactId>mvnd-daemon</artifactId>
-        <version>${project.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>org.apache.maven.daemon</groupId>
-        <artifactId>mvnd-helper-agent</artifactId>
-        <version>${project.version}</version>
+        <groupId>org.assertj</groupId>
+        <artifactId>assertj-core</artifactId>
+        <version>${assertj.version}</version>
```
mthmulders, line 145: Can we add `<scope>test</scope>` here?

agent/pom.xml
gnodet: That's right, it needs to be fixed.
However, I think we did set up a redirection and `https://maven.apache.org/xsd/maven-4.1.0.xsd` will redirect to latest beta.  I can't find out where this redirect source is...

gnodet: So `http://maven.apache.org/maven-v4_0_0.xsd` also redirects to `https://maven.apache.org/xsd/maven-4.0.0.xsd` ...

cstamas: https://github.com/apache/maven-site/tree/master/content/resources/xsd

PR #1189, opened 2024-11-07
Upgrade to Maven 4.0.0-rc-1

gnodet:
This PR updates mvnd to Maven 4.0.0-rc-1


cstamas:
All fine, manually tested as well, but one problem (and related IT) fails: raw streams
```
RawStreamsTest.version:55 Output should contain PrintOut{payload='Hello'} ==> expected: <true> but was: <false>
```

IMO this is a problem in maven and not mvnd.

.mvn/maven.config
```diff
@@ -1,3 +1,4 @@
 -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
 -Dmaven.wagon.http.retryHandler.requestSentEnabled=true
 -Dmaven.wagon.http.retryHandler.count=10
+-Daether.transport.jdk.httpVersion=HTTP_1_1
```
cstamas, line 4: This can go away now with Resolver 2.0.4 (mvn 4-beta-6)

cstamas: In fact, all of these can go away, wagon is not used anymore, is it?

gnodet:
Fixes #1203 and #1204.

daemon/src/main/java/org/apache/maven/cli/DaemonMavenInvoker.java
```diff
@@ -18,68 +18,60 @@
...
+                    builder.systemOutput(TerminalBuilder.SystemOutput.ForcedSysOut);
+                    builder.provider(TerminalBuilder.PROP_PROVIDER_EXEC);
+                    if (context.coloredOutput != null) {
+                        builder.color(context.coloredOutput);
+                    }
+                },
+                terminal -> doConfigureWithTerminal(context, terminal));
+        context.terminal = MessageUtils.getTerminal();
+        context.closeables.add(MessageUtils::systemUninstall);
+        MessageUtils.registerShutdownHook();
```
cstamas, line 62: We should remove this from here, and install it **only once** in LookupInvoker maybe?

daemon/src/main/java/org/apache/maven/cli/DaemonMavenOptions.java
```diff
@@ -24,7 +24,6 @@
 import org.apache.maven.api.cli.mvn.MavenOptions;
 
 public interface DaemonMavenOptions extends MavenOptions {
-
```
cstamas, line 27: Am unsure about the existence of this class. DaemonMavenOptions was introduced to add new CLI options like "raw streams", but today, no new option in mvnd exists...

daemon/src/main/java/org/apache/maven/cli/DaemonMavenInvoker.java
gnodet: Maven 3 did not create a `Terminal`, so this was needed. But now it should go away.

cstamas: I'd leave all this as is, as shutdown hook is also removed on uninstall, hence IMO we are ok. Also, mvnd creates "exec" terminal while mvn not.

We can further simplify later.

daemon/src/main/java/org/apache/maven/cli/DaemonMavenOptions.java
gnodet: That's kind expected, as mvnd specific options (such as `--stop`) are handled differently at the very beginning, or via system properties.

PR #1250, opened 2025-01-28
added missing override annotations

Alvari-Kupari:
This helps with readability and complies with codestyle conventions.


cstamas:
You reformatted as well. Did you try the build locally? We use spotless for formatting...

PR #1281, opened 2025-03-06
Make DaemonRegistry concurrency-safe

oehme:
The previous implementation had two issues:

The memory-mapped file was created in the constructor and ignored file expansions by other
threads or processes. It would then overwrite the file with its own, smaller size buffer.
Replaced the memory mapping with a regular InputStream/OutputStream approach. Memory mapping
is not very useful in this scenario anyway, since the file is small and infrequently read and
we always read/write the entire file.

The tryLock method was broken and only tried locking once. If the lock could not be acquired,
it returned null and the subsequent code then ignored that fact and proceeded without locking.
I've moved the retry loop and failure handling into the lock method, which gives it the intended
behavior of waiting for a lock for 20s and then failing.

The test expectation for the "big registry" test has changed, since we no longer
manually increase the file size to multiples of two, but instead simply write out
exactly the bytes we need.

I droped the String size requirement, since DataInputStream handles arbitrarly
long Strings. I also changed the exception handling to consistently do recovery
on any kind of exception while reading/writing, since it is reasonable to assume
that any such exception indicates a broken registry file. I've also adjusted the
user-facing message to only tell them to delete the file if recovery fails.

### DIFF
--- a/common/src/main/java/org/mvndaemon/mvnd/common/DaemonRegistry.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/DaemonRegistry.java
@@ -18,15 +18,15 @@
  */
 package org.mvndaemon.mvnd.common;
 
+import java.io.BufferedInputStream;
+import java.io.BufferedOutputStream;
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
-import java.nio.BufferOverflowException;
-import java.nio.BufferUnderflowException;
-import java.nio.ByteBuffer;
-import java.nio.MappedByteBuffer;
+import java.nio.channels.Channels;
 import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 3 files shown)


gnodet:
Awesome work !

oehme:
Those mac/windows failures indicate there's still something wrong - I'll rework the test to use multiple processes instead of only multiple threads. Hopefully that'll uncover the problem.

oehme:
Turns out the real issue was in the `tryLock` method, which just silently returned `null` when the file couldn't be locked. Also today I learned that `try-with-resources` will happily accept `null`, so this was never spotted. The end effect was that there was no concurrency control, all processes would access the file without coordination.

I've added a process-based test and fixed the tryLock method so that it does the retry loop until it gets a non-null lock or the timer runs out.

In fact, I could revert all the other changes, the tryLock turned out to be the only true issue. To me, DataInputStream/DataOutputStream seems more straightforward than the memory mapped buffer. Whatever you prefer.

oehme:
Sorry I had to edit my message above, you probably responded to the first version where I had doubted the memory mapping change. The memory mapping was the second part of the problem after the file locking. It only mapped size() bytes in the constructor and completely ignored if other threads or processes expanded it. 

Even if we fixed that by checking the size and remapping the file, I don't see much value. This file is small and infrequently read. For reference, Gradle also rereads it's registry rather than memory mapping

common/src/main/java/org/mvndaemon/mvnd/common/DaemonRegistry.java
```diff
@@ -18,15 +18,11 @@
  */
 package org.mvndaemon.mvnd.common;
 
-import java.io.IOException;
+import java.io.*;
```
cstamas, line 21: expand this please

cstamas:
Cool beans! Nice and thanks!

gnodet (approved):

cstamas (approved):

oehme:
Added the thread-based test back and did some final cleanups, this should be good to go now.

PR #1282, opened 2025-03-06
Bug: extension handling

### LINKED ISSUE #1280: Daemon CLI parser does not detect extension conflicts
### Affected version

2.0-SNAPSHOT

### Bug description

mvnd overrides the extension parsing code from [maven-core](https://github.com/apache/maven/blob/c414d8a5da8f2a12eecbd79ac31237b6ffe4c1c5/impl/maven-cli/src/main/java/org/apache/maven/cling/invoker/BaseParser.java#L458) and thus misses the detection of duplicate extensions

cstamas:
mvnd still had mvn3 like code, just drop it and let base parser do the work. All the mvnd needs is filtering.

mvnd re extensions was still on mvn3 level, as it supported only one extnsions source. But in fact, the problem was that it override methods that were not needed to be overrridden.

Fixes #1280

### DIFF
--- a/client/src/main/java/org/mvndaemon/mvnd/client/DaemonParameters.java
+++ b/client/src/main/java/org/mvndaemon/mvnd/client/DaemonParameters.java
@@ -450,12 +450,6 @@ public class DaemonParameters {
         if (env == Environment.MVND_EXT_CLASSPATH) {
             List<String> cp = parseExtClasspath(userHome());
             return String.join(",", cp);
-        } else if (env == Environment.MVND_CORE_EXTENSIONS_FILE_PATH) {
-            try {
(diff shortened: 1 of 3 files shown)


gnodet (approved):

cstamas:
@oehme Any remark to this one?

oehme:
Ran it through our tests and can confirm that the behavior is now like regular Maven 4. Also found no new issues :+1:

cstamas:
Thanks, will merge once main branch CI settles

cstamas:
Argh, master CI mac went read, again connectivity issue
https://github.com/orgs/community/discussions/145314

cstamas:
We may want Mimir setup here as well? 
https://github.com/apache/maven/pull/2144

PR #1283, opened 2025-03-07
[experiment] Mimir

cstamas:
Use Maveniverse Mimir to cache central artifacts. This makes caching way simpler as well. Update Maven used to build to Maven 4.0.0-rc-3 as well.


cstamas:
Once rc-3 is release, all we need is just up the mvn used to build this (from 3.9.9) to rc-3, and then move extensions.xml to user home (instead of project .mvn/extensions.xml) and it should work, as I have same setup locally and ITs do pick up extensions from my home .m2

gnodet (approved):

PR #1285, opened 2025-03-08
Adding mvnd for linux on docker on a mac.

flyhard:
I am trying to use mvnd in a devcontainer on a mac. There is no binary available to install. It is also missing from SDKMAN!

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -36,6 +36,12 @@ jobs:
       matrix:
         # binaries wanted: linux amd64, mac intel, mac M1, windows x86
         os: [ ubuntu-latest, macos-13, macos-14, windows-latest ]
+        arch: [ x64, arm64 ]
+        exclude:
+          - os: macos-13
+            arch: arm64
+          - os: windows-latest
+            arch: arm64
     runs-on: ${{ matrix.os }}
 
     steps:


cstamas:
This needs more, look at other workflow files, like release...

flyhard:
> The release publishing scripts for [homebrew](https://github.com/apache/maven-mvnd/blob/master/build/publish-on-homebrew.sh) and [sdkman](https://github.com/apache/maven-mvnd/blob/master/build/publish-on-sdkman.sh) have to be adjusted as well.

I'll have a look

flyhard:
Should the new flow only work for v2 of mvnd? or even for 1.0?

cstamas:
Just keep on master (2.x) for now.

flyhard:
Please have a look at it again to see if it works correctly now. I have this as accompanying PR: https://github.com/mvndaemon/homebrew-mvnd/pull/17

.github/workflows/release.yaml
```diff
@@ -35,7 +35,13 @@ jobs:
       fail-fast: true
       matrix:
         # binaries wanted: linux amd64, mac intel, mac M1, windows x86
-        os: [ ubuntu-latest, macos-13, macos-14, windows-latest ]
+        os: [ ubuntu-latest, macos-latest, macos-13, macos-14, windows-latest ]
```
gnodet, line 38: Unless I'm wrong, `macos-14` and `macos-15` have the same architecture, so while I'm fine dual testing those two platforms, having separate distributions makes no sense imho.

flyhard: so I guess, drop the macos-14 should be fine?

.github/workflows/release.yaml
```diff
@@ -35,7 +35,13 @@ jobs:
...
+        arch: [ x64, arm64 ]
+        exclude:
+          - os: macos-13
+            arch: arm64
+          - os: windows-latest
```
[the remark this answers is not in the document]
flyhard: sounds fine to me

.github/workflows/release.yaml
```diff
@@ -35,7 +35,13 @@ jobs:
       fail-fast: true
       matrix:
         # binaries wanted: linux amd64, mac intel, mac M1, windows x86
-        os: [ ubuntu-latest, macos-13, macos-14, windows-latest ]
+        os: [ ubuntu-latest, macos-latest, macos-13, windows-latest ]
+        arch: [ x64, arm64 ]
+        exclude:
+          - os: [ macos-13, macos-14 ]
```
gnodet, line 41:
```suggestion
          - os: [ macos-13 ]
```

flyhard:
I have tried to incorporate your changes. Does it look ok to you now?

build/publish-on-homebrew.sh
```diff
@@ -34,6 +34,8 @@ darwinArmZipUrl="https://downloads.apache.org/maven/mvnd/${VERSION}/maven-mvnd-$
 darwinArmSha256="$(curl -L --silent "${darwinArmZipUrl}.sha256")"
 linuxZipUrl="https://downloads.apache.org/maven/mvnd/${VERSION}/maven-mvnd-${VERSION}-linux-amd64.zip"
```
gnodet, line 35: maybe add the `Amd` prefix to keep things aligned ? similar to what we have for `darwin`

flyhard: I have fixed that. Please review again :-)

gnodet (approved):

flyhard:
Is there something I need to do before this can be merged?

cstamas (approved):

cstamas:
Seems something is still needed here... there is a name conflict?
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run

PR #1287, opened 2025-03-10
Upgrade to Nisse extension

gnodet:


pom.xml
```diff
@@ -654,6 +619,44 @@
...
+                <goals>
+                  <goal>format</goal>
+                </goals>
+                <phase>process-sources</phase>
+              </execution>
+            </executions>
+          </plugin>
+          <plugin>
+            <groupId>net.revelc.code</groupId>
+            <artifactId>impsort-maven-plugin</artifactId>
```
cstamas, line 649: pomsort is enabled in spotless, is this really needed also?

gnodet: Right, not sure how that managed to get into this commit, as I haven't added those explicitely.  I must have switched between branches.

.mvn/extensions.xml
```diff
@@ -1,4 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <extensions>
+    <extension>
+        <groupId>eu.maveniverse.maven.nisse</groupId>
+        <artifactId>extension</artifactId>
+        <version>0.4.0</version>
```
cstamas, line 7: 0.4.1 is released

cstamas (approved):

Issue #1288, opened 2025-03-11
Maven Daemon site build fails due to wrong skin


ascheman:
### Affected version

Latest build

### Bug description

When building the site for Maven Daemon I run into a build problem, since the project uses Maven Site Plugin in version 3.21.0 but Maven Fluido Skin (inherited from the parent) in version 1.12.0 which [is incompatible](https://maven.apache.org/skins/index.html).

## Background

* Clone: `git clone https://github.com/apache/maven-mvnd`
* Build: `cd maven-mvnd && mvn site` (using Maven 3.9.9)
* Error: `[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.21.0:site (default-site) on project mvnd: Failed to create context for skin: Cannot use skin: has [1.11.1,2.0.0-M1) Doxia Sitetools prerequisite, but current is 2.0.0 -> [Help 1]`

Found this as part of support-and-care/maven-support-and-care#77 to make the Maven Sources reactor build again.

ascheman:
See this problem in my forked build: https://github.com/support-and-care/maven-mvnd/actions/runs/13782628403/job/38543517230

PR #1289, opened 2025-03-11
Build Site and fix 1288

ascheman:
Besides a fix for #1288 the change also contains 

* a respective CI build (`mvn ... site`) to detect such problems early in the future
* a simplification for GH CI triggers (build on each push, not only master and PRs)

The latter leads to CI builds for each development change (which is a major purpose of an automatic build and helps in particular for forks/new contributors who cannot issue a automatic build by creating a PR). Having said that, the respective commit could be dropped if there is a good reason not to follow this practice.

### DIFF
--- a/.github/workflows/early-access.yaml
+++ b/.github/workflows/early-access.yaml
@@ -20,9 +20,6 @@ name: Early Access
 # trigger on push to branches and PR
 on:
   push:
-    branches:
-      - master
-      - mvnd-1.x
   pull_request:
 
 env:
--- a/.github/workflows/early-access.yaml
+++ b/.github/workflows/early-access.yaml
@@ -47,7 +47,7 @@ jobs:
           distribution: 'temurin'
 
       - name: 'Run default (non-native) build'
-        run: ./mvnw verify -Dmrm=false -V -B -ntp -e -s .mvn/release-settings.xml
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -95,6 +95,8 @@
     <version.plexus-xml>4.0.4</version.plexus-xml>
     <jakarta.inject.version>2.0.1</jakarta.inject.version>
 
+    <version.maven-fluido-skin>2.0.1</version.maven-fluido-skin>
```
cstamas, line 98: Isn't 2.1.0 just released? Why 2.0.1 exactly?

.github/workflows/early-access.yaml
```diff
@@ -47,7 +44,7 @@ jobs:
           distribution: 'temurin'
 
       - name: 'Run default (non-native) build'
-        run: ./mvnw verify -Dmrm=false -V -B -ntp -e -s .mvn/release-settings.xml
+        run: ./mvnw verify site -Dmrm=false -V -B -ntp -e -s .mvn/release-settings.xml
```
cstamas, line 47: I'd not interleave site into this one... @gnodet ?

gnodet: Yes, maybe another job would be better specifically for the site.

ascheman: Moved it to another Job

gnodet (approved):
LGTM

PR #1303, opened 2025-03-27
Make prompting work in the daemon

### LINKED ISSUE #1292: Prompter does not work
### Affected version

2.0.0-rc-3

### Bug description

When I try to prompt for user input, the build just hangs

```
   java.lang.Thread.State: TIMED_WAITING (on object monitor)
	at java.lang.Object.wait0(java.base@21.0.6/Native Method)
	- waiting on <no object reference available>
	at java.lang.Object.wait(java.base@21.0.6/Object.java:366)
	at org.jline.utils.NonBlockingPumpInputStream.wait(NonBlockingPumpInputStream.java:53)
	at org.jline.utils.NonBlockingPumpInputStream.read(NonBlockingPumpInputStream.java:90)
	- locked <0x000000041f9b5d38> (a org.jline.utils.NonBlockingPumpInputStream)
	at org.jline.utils.NonBlockingInputStream.read(NonBlockingInputStream.java:62)
	at org.jline.utils.NonBlocking$NonBlockingInputStreamReader.read(NonBlocking.java:157)
	at org.jline.utils.NonBlockingReader.read(NonBlockingReader.java:56)
	at org.jline.keymap.BindingReader.readCharacter(BindingReader.java:159)
	at org.jline.keymap.BindingReader.readBinding(BindingReader.java:109)
	at org.jline.keymap.BindingReader.readBinding(BindingReader.java:61)
	at org.jline.reader.impl.LineReaderImpl.doReadBinding(LineReaderImpl.java:973)
	at org.jline.reader.impl.LineReaderImpl.readBinding(LineReaderImpl.java:1006)
	at org.jline.reader.impl.LineReaderImpl.readLine(LineReaderImpl.java:689)
	at org.jline.reader.impl.LineReaderImpl.readLine(LineReaderImpl.java:552)
	at org.jline.reader.impl.LineReaderImpl.readLine(LineReaderImpl.java:523)
	at org.apache.maven.jline.DefaultPrompter.doPrompt(DefaultPrompter.java:82)
	at org.apache.maven.jline.DefaultPrompter.doPrompt(DefaultPrompter.java:105)
	at org.apache.maven.jline.DefaultPrompter.prompt(DefaultPrompter.java:39)
	at org.apache.maven.api.services.Prompter.prompt(Prompter.java:44)
        <our plugin here>
```

Looks like the `DaemonPrompter` is never used and instead it is using the regular prompter that is waiting for input on the daemon server side instead of the client side.

I think this is because `DaemonPrompter` implements the wrong interface: `org.codehaus.plexus.components.interactivity.Prompter` vs `org.apache.maven.api.services.Prompter`

oehme:
The daemon was implementing the legacy plexus.interactivity Prompter interface, which Maven 4 no longer calls. Instead, it needs to implement the new maven.api Prompter interface. This also simplifies the code, since the legacy prompter handling is now done by Maven core and it delegates to the new interface if an old plugin uses it.

Also, prompting was skipped on dumb terminals for some reason. This meant that prompting could not be tested with automated tests using ProcessBuilder. It also meant that piping input into mvnd on the terminal wouldn't work. This made it inconsistent with regular Maven, which supports prompting on a dumb terminal. I've reactivated it, which makes the prompting integration tests for our Maven extension pass again.

There was also a missing console flush before sending the response to the daemon, which could lead to inconsistent output, where the newline was sometimes printed before and sometimes after the next line from the daemon.

Fixes #1292

### DIFF
--- a/common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalInputHandler.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalInputHandler.java
@@ -178,6 +178,7 @@ public class TerminalInputHandler implements AutoCloseable {
                 break;
             } else if (c == '\n' || c == '\r') {
                 terminal.writer().println();
+                terminal.writer().flush();
                 daemonDispatch.accept(prompt.response(sb.toString()));
                 break;
             } else if (c == 127) { // Backspace
--- a/common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalOutput.java
@@ -328,10 +328,6 @@ public class TerminalOutput implements ClientOutput {
             }
             case Message.PROMPT: {
                 Message.Prompt prompt = (Message.Prompt) entry;
-                if (dumb) {
-                    terminal.writer().println("");
-                    break;
-                }
                 clearDisplay();
                 inputHandler.requestPrompt(prompt);
                 break;
--- a/daemon/src/main/java/org/mvndaemon/mvnd/interactivity/DaemonPrompter.java
+++ b/daemon/src/main/java/org/mvndaemon/mvnd/interactivity/DaemonPrompter.java
@@ -26,12 +26,9 @@ import java.util.List;
 import java.util.Objects;
 import java.util.UUID;
 
+import org.apache.maven.api.services.Prompter;
+import org.apache.maven.api.services.PrompterException;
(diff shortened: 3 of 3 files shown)


oehme:
Ah, the failure is because the [legacy support in maven-core](https://github.com/apache/maven/blob/master/compat/maven-compat/src/main/java/org/apache/maven/internal/compat/interactivity/LegacyPlexusInteractivity.java#L46) is broken and doesn't do what it says in the Javadoc. It's supposed to delegate to the Maven 4 API, but instead it delegates to an implementation detail. So that'll have to be fixed as well.

cstamas (approved):

PR #1319, opened 2025-05-10
Drop buildnumber plugin, Nisse is in place

### LINKED ISSUE #1306: Can't build from source tarball
### Affected version

2.0.0-rc3

### Bug description

Downloaded the source tarball for 2.0.0-rc3 from [the Apache Maven website](https://maven.apache.org/download.cgi). Extracted and tried to build. The build fails with

```
[ERROR] Failed to execute goal org.codehaus.mojo:buildnumber-maven-plugin:3.2.1:create (default) on project mvnd-client: Cannot get the revision information from the scm repository :
[ERROR] Exception while executing SCM command.: The git log command failed: /bin/sh -c cd '/home/ubuntu/mvnd/maven-mvnd-2.0.0-rc-3/client' && 'git' 'log' '-1' '--no-merges' '--format=format:%H %aI %aE %aN' returned fatal: not a git repository (or any of the parent directories): .git
```

Building from a Git clone where I checked out the corresponding tag works just fine.

cstamas:
Just use Nisse, as it is already in place, plus it does not require git checkout either.

Fixes #1306

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -78,21 +78,6 @@
     </resources>
     <plugins>
       <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>buildnumber-maven-plugin</artifactId>
-        <configuration>
-          <locale>en_US</locale>
-        </configuration>
-        <executions>
-          <execution>
-            <goals>
-              <goal>create</goal>
-            </goals>
-            <phase>validate</phase>
-          </execution>
-        </executions>
-      </plugin>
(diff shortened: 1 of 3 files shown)


cstamas:
Nota bene: this does not solve "rebuilding from tarball", as output will still not contain revision (will be empty). Just like Maven, MavenDaemon also relies on git to provide SCM hash shown in version.

gnodet (approved):

PR #1412, opened 2025-09-06
Try to downgrade Win

cstamas:


mbien:
yey, green!

cstamas:
Probably yes, but on a longer run we should solve this.

cstamas:
No, in case of mvnd-1.x it was this:
https://github.com/apache/maven-mvnd/blob/mvnd-1.x/common/src/main/java/org/mvndaemon/mvnd/common/OsUtils.java#L63

gnodet:
> No, in case of mvnd-1.x it was this: https://github.com/apache/maven-mvnd/blob/mvnd-1.x/common/src/main/java/org/mvndaemon/mvnd/common/OsUtils.java#L63
> 
> Well, maybe yes? But the code in question is "prod" codebase.

See https://github.com/apache/maven-mvnd/pull/1414

PR #1416, opened 2025-09-08
Fix workflow

cstamas:
On master it doubles CI jobs for all in-repo
PRs (as it is push but also PR).


cstamas:
Seems this was already present but removed by this:
https://github.com/apache/maven-mvnd/commit/cb2c732201b5170d7c99fcb921abda76d205c264

@ascheman why was removal needed?

PR #1428, opened 2025-09-12
Produce Linux arm64 binaries as well

cstamas:


.github/workflows/early-access.yaml
```diff
@@ -156,6 +157,27 @@ jobs:
...
 
+      - name: '(arm64) Patch GraalVM libs for only requiring glibc 2.12'
+        if: ${{ env.OS == 'linux' && env.ARCH == 'aarch64' }}
+        shell: bash
+        run: |
+          mkdir -p client/target/graalvm-libs-for-glibc-2.12
+
+          : patch common libraries
+          ( find "$GRAALVM_HOME/lib/static/linux-aarch64/glibc" -name '*.a'
+            ls -1 /lib/aarch64-linux-gnu/libz.a
```
ls-urs-keller, line 168: Appreciate this PR. You could use something like $(arch) instead of aarch64 or x86_64, this way this code can be the same as the one for amd64. 
```bash
$ docker run --platform linux/arm64 -ti alpine arch
aarch64
$ docker run --platform linux/amd64 -ti alpine arch
x86_64
```

PR #1527, opened 2026-02-05
:construction_worker: (feat): Add ARM64 Builds

KrisJohnstone:
Added support for ARM64 builds using glibc 2.17.


cstamas:
Yes, this looks more in line with what summoned @gzm55 proposed:
* leave linux/amd64 "as is" (as it was)
* make linux/arm64 use glibc 2.28

cstamas:
Will leave it here. Seems patching is not correct? Also, keep in mind to keep in sync the two workflows (early preview and release).

KrisJohnstone:
First laptop was DOA. Back on this now :)

PR #1558, opened 2026-03-17
Fix flaky awaitIdle in integration tests

gnodet:
## Summary

- Add 100ms sleep between polling iterations in `TestRegistry.awaitIdle()` to avoid tight busy-loop spinning on file lock acquisition
- Increase timeout from 5s to 30s to accommodate slow CI environments

## Problem

`StopStatusTest.stopStatus` (and other tests using `awaitIdle`) intermittently fail on macOS ARM64 CI runners with:
```
java.lang.AssertionError: Daemon <id> should have become idle within 5000
```

Root cause: `awaitIdle()` was a tight busy-loop calling `getAll()` on every iteration. Each `getAll()` call acquires a file lock on the daemon registry, causing contention. The 5000ms timeout was too short for CI runners under load.

## Test plan

- [x] Builds successfully
- The fix reduces lock contention and provides a more generous timeout, making the test robust on slow CI environments without significantly increasing test duration (the sleep only applies while waiting)

### DIFF
--- a/integration-tests/src/test/java/org/mvndaemon/mvnd/junit/TestRegistry.java
+++ b/integration-tests/src/test/java/org/mvndaemon/mvnd/junit/TestRegistry.java
@@ -72,7 +72,7 @@ public class TestRegistry extends DaemonRegistry {
      * @throws AssertionError        if the timeout is exceeded
      */
     public void awaitIdle(String daemonId) {
-        final int timeoutMs = 5000;
+        final int timeoutMs = 30_000;
         final long deadline = System.currentTimeMillis() + timeoutMs;
         while (getAll().stream()
                         .filter(di -> di.getId().equals(daemonId))
@@ -84,6 +84,12 @@ public class TestRegistry extends DaemonRegistry {
             Assertions.assertThat(deadline)
                     .withFailMessage("Daemon %s should have become idle within %d", daemonId, timeoutMs)
                     .isGreaterThan(System.currentTimeMillis());
+            try {
+                Thread.sleep(100);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
+            }
         }
     }
 }


gnodet (commented):
Clean, minimal fix for a real flaky test issue. The 100ms polling sleep matches the existing interval in `DaemonRegistry.tryLock()`, and the timeout increase from 5s to 30s is justified as a ceiling for loaded CI runners. The interrupt handling pattern is correct and the sleep placement after the deadline assertion limits worst-case overshoot to ~100ms.

LGTM — well-scoped and targeted.

slawekjaranowski (approved):

gnodet (commented):
Re-review: The new approval from slawekjaranowski (2026-07-28) confirms this is ready. No new concerns raised.

Clean, minimal fix — the 100ms polling sleep matches existing patterns in `DaemonRegistry.tryLock()`, interrupt handling follows best practices, and the 30s timeout ceiling is justified for slow CI.

LGTM.

gnodet (commented):
AI-assisted review — looks good ✅

Clean, minimal fix for a well-understood CI flakiness issue. The 100ms polling sleep and timeout increase from 5s to 30s are both sensible changes with correct interrupt handling.

The sleep is correctly placed after the deadline assertion, so the method fails immediately when the deadline has passed rather than sleeping an extra 100ms first. The interrupt handling follows best practice (re-setting the interrupt flag before re-throwing). The 6x timeout increase is justified for CI environments under load and does not affect happy-path test duration.

gnodet (commented):
Clean and correct fix for a flaky test. The 100ms sleep breaks the tight busy-loop in `awaitIdle()`, and the 30s timeout accommodates slow CI runners. Both changes are well-justified and consistent with existing patterns in the codebase (e.g., the 100ms interval in `DaemonRegistry.tryLock()`).

Looks good to merge. ✅

gnodet (commented):
**AI Review — LGTM** ✅

Clean, minimal fix for a real flaky test issue. The 100ms polling sleep and 30s timeout are both well-justified and correctly implemented.

**Key observations:**
- The fix is correct: `getAll()` calls `read()` which acquires a file lock via `tryLock()`, so the original tight loop was a hot spin on file lock acquisition. Adding 100ms sleep matches the exact polling interval used in `DaemonRegistry.tryLock()` itself.
- The sleep placement after the deadline assertion is the right ordering: the method fails immediately on timeout rather than sleeping an extra 100ms before checking.
- Interrupt handling follows best practice: re-sets the interrupt flag via `Thread.currentThread().interrupt()` before re-throwing as a RuntimeException.
- The timeout increase from 5s to 30s is reasonable as a ceiling for loaded CI environments; it does not slow down the happy path since the loop exits as soon as the daemon becomes idle.

PR #1559, opened 2026-03-17
Add zsh support to completion script

gnodet:
## Summary

- Add zsh support to the existing `mvnd-bash-completion.bash` script (auto-detects shell type)
- Accept `zsh` as valid `--completion` argument in `Completion.java`
- Update README.adoc with zsh instructions

Supersedes #1514 by consolidating the zsh compatibility into the single existing script rather than adding a separate file. This avoids duplication and maintenance burden.

## Approach

The script auto-detects bash vs zsh and adapts three shell-specific behaviors:
- `function_exists`: uses `typeset -f` (zsh) vs `declare -F` (bash)
- Plugin variable discovery: uses `typeset +` (zsh) vs `compgen -v` (bash)
- Indirect variable expansion: uses `${(P)var}` (zsh) vs `${!var}` (bash)

For zsh, `bashcompinit` is loaded to provide bash-style completion builtins (`complete`, `compgen`, `COMPREPLY`). No global shell options (`setopt`) are modified.

## Credit

Based on the approach from #1514 by @janweinschenker.

## Test plan

- [x] Full build passes
- [x] Completion template regeneration via `CompletionGeneratorTest` succeeds
- Source the script in bash: works as before
- Source the script in zsh: completions work via bashcompinit layer

### DIFF
--- a/dist/src/main/distro/bin/mvnd-bash-completion.bash
+++ b/dist/src/main/distro/bin/mvnd-bash-completion.bash
@@ -17,10 +17,26 @@
 # Adapted from https://github.com/juven/maven-bash-completion/blob/master/bash_completion.bash by Juven Xu and others
 # under Apache License Version 2.0
 
+# Detect shell type and set up zsh compatibility if needed
+if [ -n "$ZSH_VERSION" ]; then
+    __MVND_SHELL="zsh"
+    # Load bashcompinit for bash-style completions in zsh
+    autoload -Uz bashcompinit 2>/dev/null && bashcompinit
+    # Define COMP_WORDBREAKS if not set (bash sets this automatically)
+    [[ -z "$COMP_WORDBREAKS" ]] && COMP_WORDBREAKS=$' \t\n"\'><=;|&(:'
+else
+    __MVND_SHELL="bash"
+fi
+
 function_exists()
 {
-    declare -F $1 > /dev/null
-    return $?
+    if [ "$__MVND_SHELL" = "zsh" ]; then
+        typeset -f $1 > /dev/null 2>&1
+        return $?
+    else
+        declare -F $1 > /dev/null
+        return $?
+    fi
 }
 
 # This function can be used to access a tokenized list of words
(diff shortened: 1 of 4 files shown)


gnodet (commented):
Clean, well-structured PR that adds zsh support to the existing bash completion script. The three shell-specific adaptations (function detection via `typeset -f`, variable discovery via `typeset +`, indirect expansion via `${(P)var}`) are correctly implemented using standard zsh idioms, and the `bashcompinit` approach is simpler than the superseded PR #1514.

Both copies of the completion script (dist and test template) are properly synchronized.

One minor suggestion: consider adding a `completionZsh()` test in `CompletionNativeIT.java` to exercise the new `--completion zsh` code path. Since both shells return the same script, the assertions can be identical to the bash test, but it would guard against regressions in the argument validation.

LGTM.

gnodet (commented):
AI-assisted review — looks good ✅

Clean, well-scoped PR that adds zsh compatibility to the existing bash completion script via shell auto-detection. The three zsh-specific adaptations are correctly implemented using standard idioms, and the consolidation approach (single file instead of a duplicate) is superior to the superseded PR #1514.

Good design choices: keeping a single auto-detecting script avoids duplication, not calling `compinit` unconditionally avoids shell startup slowdown, and the `__MVND_SHELL` double-underscore prefix is a good convention to avoid variable collisions. Both copies of the completion script (dist and test template) are correctly kept in sync.

dist/src/main/distro/bin/mvnd-bash-completion.bash
```diff
@@ -17,10 +17,26 @@
 # Adapted from https://github.com/juven/maven-bash-completion/blob/master/bash_completion.bash by Juven Xu and others
 # under Apache License Version 2.0
 
+# Detect shell type and set up zsh compatibility if needed
+if [ -n "$ZSH_VERSION" ]; then
+    __MVND_SHELL="zsh"
+    # Load bashcompinit for bash-style completions in zsh
+    autoload -Uz bashcompinit 2>/dev/null && bashcompinit
+    # Define COMP_WORDBREAKS if not set (bash sets this automatically)
```
gnodet, line 25: **Suggestion (medium):** The zsh documentation states that `bashcompinit` must be called after `compinit` has been loaded. This script calls `autoload -Uz bashcompinit && bashcompinit` without first ensuring `compinit` is initialized. Most zsh users will already have `compinit` loaded via their framework (oh-my-zsh, prezto, etc.) or `.zshrc`, but users with a minimal zsh configuration will get silent failure — the `2>/dev/null` suppresses the error and completions simply won't register.

Consider adding `compinit` initialization before `bashcompinit`:
```suggestion
    __MVND_SHELL="zsh"
    # Load zsh completion system if not already initialized
    autoload -Uz compinit 2>/dev/null && compinit -u 2>/dev/null
    # Load bashcompinit for bash-style completions in zsh
    autoload -Uz bashcompinit 2>/dev/null && bashcompinit
```

Alternatively, document the `compinit` prerequisite in the README.

gnodet (commented):
Clean, well-structured PR that adds zsh support to the existing bash completion script. The approach of consolidating into a single script (instead of a separate file as in the superseded PR #1514) is the right design choice. No breaking changes to existing bash behavior.

The three zsh-specific adaptations (`typeset -f` for function checks, `typeset +` for variable listing, `${(P)var}` for indirect expansion) are the canonical approaches for bash/zsh dual scripts. Good decision to skip `setopt` calls that PR #1514 included.

One minor suggestion below about the `compinit` prerequisite for minimal zsh setups.

gnodet (commented):
**AI Review — LGTM** ✅

Clean, well-scoped PR that correctly adds zsh support to the existing bash completion script via shell auto-detection and three canonical zsh idiom substitutions. Backwards compatibility with bash is fully preserved.

**Key strengths:**
- The design choice to consolidate bash and zsh into a single auto-detecting script (rather than a separate file) avoids duplication and maintenance burden.
- The three shell-specific adaptations (`typeset -f`, `typeset +`, `${(P)var}`) are all canonical zsh idioms and correctly implemented.
- Adding `local` to the `var_name` declaration is an incidental improvement: the original code leaked `var_name` into the parent scope.
- Both copies of the completion script are properly kept in sync.

**Minor suggestions (non-blocking):**

1. **compinit guard:** The script loads `bashcompinit` but does not guard against `compinit` not having been run yet. A lightweight guard like `if ! whence -w compdef >/dev/null 2>&1; then autoload -Uz compinit && compinit -u 2>/dev/null; fi` before the `bashcompinit` call would cover edge cases.

2. **Missing zsh test:** No test exists for the `--completion zsh` code path. A `completionZsh()` test would guard against regressions in argument validation.

3. **Pre-existing unquoted `$1`:** The `$1` argument in `function_exists()` is unquoted in both branches. Not introduced by this PR, safe in practice.

PR #1648, opened 2026-06-23
Fix stdin behaviour when piping to mvnd

### LINKED ISSUE #1452: Can only read 1 byte from stdin
### Affected version

```
Apache Maven Daemon (mvnd) 2.0.0-rc-4-SNAPSHOT linux-amd64 native client (6734804eedc68f0f30aca315e842f3ec015def73)
Terminal: org.jline.terminal.impl.PosixSysTerminal with pty org.jline.terminal.impl.jni.linux.LinuxNativePty
Apache Maven 4.0.0-rc-4 (bed0f8174bf728978f86fac533aa38a9511f3872)
Maven home: /home/jay/Projects/maven-mvnd/dist/target/maven-mvnd-2.0.0-rc-4-SNAPSHOT-linux-amd64/mvn
Java version: 24.0.2, vendor: GraalVM Community, runtime: /home/jay/.local/jvm/jdk24-graalvm
Default locale: en_GB, platform encoding: UTF-8
OS name: "linux", version: "6.17.0-2-default", arch: "amd64", family: "unix"
```

### Bug description

I am looking to use mvnd as a way to run an auto-formatter in an IDE plugin, so far it looks much faster than mvn so I'm very keen to get it working.

The auto-formatter mojo tries to read the document from stdin. I have noticed this works fine in mvn but on mvnd it is only able to read 1 byte. I made the following mojo to demonstrate the issue:

```java
package sample.plugin;

import org.apache.maven.plugin.AbstractMojo;
import org.apache.maven.plugin.MojoExecutionException;
import org.apache.maven.plugins.annotations.Mojo;
import java.nio.charset.StandardCharsets;

@Mojo(name = "sayhi")
public class GreetingMojo extends AbstractMojo {
    @Override
    public void execute() throws MojoExecutionException {
        byte[] bytes = new byte[] {};

        try {
            System.out.println("Available bytes: " + System.in.available());
            bytes = System.in.readAllBytes();
        } catch (Exception ex) {
            getLog().error("Failed to read from stdin", ex);
        }

        System.out.println("Read " + bytes.length + " bytes from stdin.");
        System.out.println("Stdin content: \"" + new String(bytes, StandardCharsets.UTF_8) + "\"");
    }
}
```

If I run this with `mvn` as follows:

```
$ echo -n "Hello" | mvn sample.plugin:hello-maven-plugin:1.0-SNAPSHOT:sayhi
```

then I get the expected output:

```
[INFO] --- hello:1.0-SNAPSHOT:sayhi (default-cli) @ hello-maven-plugin ---
[INFO] [stdout] Available bytes: 5
[INFO] [stdout] Read 5 bytes from stdin.
[INFO] [stdout] Stdin content: "Hello"
```

However if I use `mvnd`:

```
$ echo -n "Hello" | mvnd sample.plugin:hello-maven-plugin:1.0-SNAPSHOT:sayhi
```

then I get this output:

```
[INFO] --- hello:1.0-SNAPSHOT:sayhi (default-cli) @ hello-maven-plugin ---
[INFO] [stdout] Available bytes: 0
[INFO] [stdout] Read 1 bytes from stdin.
[INFO] [stdout] Stdin content: "H"
```

or sometimes I get `[INFO] [stdout] Stdin content: ""`. I have found if I ignore `System.in.available()` and keep trying to read 1 character at a time I can eventually get the whole stdin content, but it would be nice if `readAllBytes` just worked.

Could this be a delay in the daemon receiving the standard input from the terminal interface? I did try adding a sleep for 1s at the start of my mojo but this didn't change the behaviour at all.

SapiensAnatis:
Closes #1452.

This PR addresses two issues with standard input handling, when invoking `mvnd` in a dumb terminal with piped input, for example:

```
echo "Hello world" | mvnd somegoal
```

The motivating scenario for this is to use Spotless with mvnd, which includes an IDE integration that uses stdin to pipe a document to be formatted and sent back on stdout instead of providing a path and relying on slower filesystem reads/writes: https://github.com/diffplug/spotless/blob/main/plugin-maven/IDE_HOOK.md

The issues are:

1. `DaemonInputStream.available()` often misleadingly reports 0 or 1, and in general can only report bytes that have already been pre-read into a buffer by the daemon
    - This PR addresses this by introducing 2 new message types: one that asks clients to call `System.in.available()` and one that contains the result of this operation.
3. If `TerminalInputHandler` is able to read to the end of stdin (as it will in a pipe scenario) it sends an EOF message _and then_ the data message. This leads to the data being ignored by `DaemonInputStream` as it sets the `eof` field to true, which causes a return from `read` with 0 bytes. 
    - This PR re-arranges the order in which these messages are sent to ensure data is sent before an EOF.

### DIFF
--- a/common/src/main/java/org/mvndaemon/mvnd/common/Message.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/Message.java
@@ -65,6 +65,8 @@ public abstract class Message {
     public static final int PRINT_ERR = 26;
     public static final int REQUEST_INPUT = 27;
     public static final int INPUT_DATA = 28;
+    public static final int REQUEST_INPUT_AVAILABLE = 29;
+    public static final int INPUT_AVAILABLE_DATA = 30;
 
     final int type;
 
@@ -122,6 +124,10 @@ public abstract class Message {
                 return RequestInput.read(input);
             case INPUT_DATA:
                 return InputData.read(input);
+            case REQUEST_INPUT_AVAILABLE:
+                return RequestInputAvailable.read(input);
+            case INPUT_AVAILABLE_DATA:
+                return InputAvailableData.read(input);
         }
         throw new IllegalStateException("Unexpected message type: " + type);
     }
@@ -1123,6 +1129,66 @@ public abstract class Message {
         }
     }
 
+    public static class RequestInputAvailable extends Message {
+
+        private final String projectId;
+
+        public static RequestInputAvailable read(DataInputStream input) throws IOException {
+            String projectId = readUTF(input);
+            return new RequestInputAvailable(projectId);
+        }
+
+        RequestInputAvailable(String projectId) {
+            super(REQUEST_INPUT_AVAILABLE);
+            this.projectId = projectId;
+        }
+
+        public String getProjectId() {
+            return projectId;
+        }
+
+        @Override
+        public String toString() {
+            return "RequestInputAvailable{" + "projectId='" + projectId + "'}";
+        }
+
+        @Override
+        public void write(DataOutputStream output) throws IOException {
+            super.write(output);
+            writeUTF(output, projectId);
+        }
+    }
+
+    public static class InputAvailableData extends Message {
+
+        private final int bytesAvailable;
+
+        public static InputAvailableData read(DataInputStream input) throws IOException {
+            int bytesAvailable = input.readInt();
+            return new InputAvailableData(bytesAvailable);
+        }
+
+        InputAvailableData(int bytesAvailable) {
+            super(INPUT_AVAILABLE_DATA);
+            this.bytesAvailable = bytesAvailable;
+        }
+
+        public int getBytesAvailable() {
+            return bytesAvailable;
+        }
+
+        @Override
+        public String toString() {
+            return "InputAvailableData{" + "bytesAvailable='" + bytesAvailable + "'}";
+        }
+
+        @Override
+        public void write(DataOutputStream output) throws IOException {
+            super.write(output);
+            output.writeInt(bytesAvailable);
+        }
+    }
+
     public int getType() {
         return type;
     }
@@ -1141,6 +1207,14 @@ public abstract class Message {
 
     public static InputData inputEof() {
         return new InputData(null);
+    }
+
+    public static RequestInputAvailable requestInputAvailable(String projectId) {
+        return new RequestInputAvailable(projectId);
+    }
+
+    public static InputAvailableData inputAvailableData(int bytesAvailable) {
+        return new InputAvailableData(bytesAvailable);
     }
 
     public static StringMessage out(String message) {
--- a/common/src/main/java/org/mvndaemon/mvnd/common/Message.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/Message.java
@@ -155,6 +155,8 @@ public abstract class Message {
             case PRINT_ERR:
             case REQUEST_INPUT:
             case INPUT_DATA:
+            case REQUEST_INPUT_AVAILABLE:
+            case INPUT_AVAILABLE_DATA:
                 return 2;
             case PROJECT_STARTED:
                 return 3;
--- a/common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalInputHandler.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/logging/TerminalInputHandler.java
@@ -218,6 +218,9 @@ public class TerminalInputHandler implements AutoCloseable {
     }
 
     private void handleAvailableLength() throws IOException {
+        if (daemonReceive == null) {
+            return;
+        }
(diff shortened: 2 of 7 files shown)


gnodet (changes requested):
Good work addressing real stdin handling bugs (EOF ordering and `available()` accuracy). The core approach — using a `ReentrantLock` with separate `Condition`s for data vs. availability — is clean.

However, the new message types will crash at runtime:

### 1. Missing `getClassOrder()` cases (high)

`REQUEST_INPUT_AVAILABLE` (29) and `INPUT_AVAILABLE_DATA` (30) are not handled in the `getClassOrder()` switch in `Message.java`, which throws `IllegalStateException` on its default branch. The daemon's `sendQueue` is a `PriorityBlockingQueue` using `Message.getMessageComparator()` (calls `getClassOrder`), so inserting a `RequestInputAvailable` message via the `requestAvailable` callback will crash when the comparator encounters it.

Fix: add them alongside the existing `REQUEST_INPUT` and `INPUT_DATA` cases:
```java
case REQUEST_INPUT:
case INPUT_DATA:
case REQUEST_INPUT_AVAILABLE:
case INPUT_AVAILABLE_DATA:
    return 2;
```

### 2. Missing null guard in `handleAvailableLength()` (medium)

Both `handleProjectInput()` and `handleControlKeys()` check `if (daemonReceive == null) { return; }` before calling `daemonReceive.accept()`. The new `handleAvailableLength()` omits this guard. Since `daemonReceive` is set lazily via `setDaemonReceive()`, there's a race window.

### 3. Unused imports in `CountBytesMojo.java` (low)

`BufferedReader`, `File`, `FileWriter`, and `Scanner` are imported but not used.

SapiensAnatis:
Thanks for the review @gnodet, I've applied the fixes. Good catch on the switch statement, it's a shame we're switching over an `int` and can't get the compiler to check it for us with a switch expression.

gnodet (approved):
## Re-review: Previous Findings Resolved

All three findings from the previous review have been correctly addressed in commit 0bb527a0:

1. **Missing `getClassOrder()` cases (HIGH)** — `REQUEST_INPUT_AVAILABLE` and `INPUT_AVAILABLE_DATA` now return priority 2 alongside the existing `REQUEST_INPUT`/`INPUT_DATA` cases. No more runtime crash risk.
2. **Missing null guard in `handleAvailableLength()`** — now has the null guard for `daemonReceive` matching the pattern used by `handleProjectInput()` and `handleControlKeys()`.
3. **Unused imports in `CountBytesMojo.java`** — `BufferedReader`, `File`, `FileWriter`, `Scanner`, `Parameter` have been removed.

The overall stdin piping implementation is sound:
- The `Lock`/`Condition` threading model is correct — separate `Condition` variables for data (`datasReadyCondition`) and availability (`availableReadyCondition`) allow both `available()` and `read()` to wait independently.
- The EOF ordering fix is correct: when end-of-stream is reached mid-read, accumulated data is sent via `inputResponse()` before sending `inputEof()`, preventing the race where `DaemonInputStream` would set `eof=true` before buffering the data.
- The 1-second timeout on `available()` provides a safe fallback if the client does not respond.
- The new integration test (`installPluginAndCount`) provides good coverage for the `available()`/`read()` round-trip.

Nice work addressing the feedback!

SapiensAnatis:
@gnodet Thanks for the re-review! What are the next steps to getting this merged? Do we need a second reviewer?

GitHub actions aren't running either, looks like a permissions issue.

gnodet (approved):
Well-implemented fix for two real stdin piping bugs (EOF ordering and `available()` accuracy) with sound concurrency, good test coverage, and all previous review findings properly addressed.

The core fix is correct: sending data before EOF in `handleProjectInput()` eliminates the race where `DaemonInputStream` would set `eof=true` before buffering pending data. The `Lock`/`Condition` threading model is sound — separate `Condition` variables for data and availability allow `available()` and `read()` to wait independently. The 1-second timeout on the `available()` round-trip is a good safety net.

SapiensAnatis:
Um, I think you already approved above with the same comments. I'm more asking about project processes, it looks like anyone with write access can merge it in but we probably want the CI to run first?

gnodet (approved):
Re-review: no new commits since 0bb527a0. The code remains sound — all three prior findings were addressed, the concurrency model (Lock/Condition) is correct and deadlock-free, and the EOF ordering fix eliminates the race described in #1452.

Re: the author's process question — CI (GitHub Actions) is not running on this PR, likely because fork PRs need maintainer approval to trigger workflows. A maintainer should approve the CI run, verify it passes, and merge.

SapiensAnatis:
@slachiewicz Hello, sorry for the ping, but I saw that you merged the most recent community PR - would you be able to kick off CI on this one? Thank you :)

slachiewicz (approved):

SapiensAnatis:
Thank you!

Issue #877, opened 2023-08-14
Failed to load native library:jansi-2.4.0-d705f3ccea969984-libjansi.so


nick-ling:
There is a problem with the maven configuration environment in Synology

```shell
limbo@Limbo-NAS:/opt$ java -version

java version "1.8.0_202"
Java(TM) SE Runtime Environment (build 1.8.0_202-b08)
Java HotSpot(TM) 64-Bit Server VM (build 25.202-b08, mixed mode)

limbo@Limbo-NAS:/opt$ mvn -v

Failed to load native library:jansi-2.4.0-320a2697b8944d1c-libjansi.so. The native library file at /tmp/jansi-2.4.0-320a2697b8944d1c-libjansi.so is not executable, make sure that the directory is mounted on a partition without the noexec flag, or set the jansi.tmpdir system property to point to a proper location.  osinfo: Linux/x86_64
java.lang.UnsatisfiedLinkError: /tmp/jansi-2.4.0-320a2697b8944d1c-libjansi.so: /tmp/jansi-2.4.0-320a2697b8944d1c-libjansi.so: failed to map segment from shared object
Apache Maven 3.8.8 (4c87b05d9aedce574290d1acc98575ed5eb6cd39)
Maven home: /opt/maven38
Java version: 1.8.0_202, vendor: Oracle Corporation, runtime: /volume1/container/common/opt/config/jdk8/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.4.302+", arch: "amd64", family: "unix"

limbo@Limbo-NAS:/opt$ 
```
> Upgrade the jdk of higher version

```shell
limbo@Limbo-NAS:/volume1/container/common/opt/config$ sudo vim /etc/profile 
limbo@Limbo-NAS:/volume1/container/common/opt/config$ source /etc/profile 
limbo@Limbo-NAS:/volume1/container/common/opt/config$ java -version

java version "17.0.8" 2023-07-18 LTS
Java(TM) SE Runtime Environment (build 17.0.8+9-LTS-211)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.8+9-LTS-211, mixed mode, sharing)

limbo@Limbo-NAS:/volume1/container/common/opt/config$ mvn -v

[...]

PR #1655, opened 2026-06-25
Fix(#1555): stale plugin descriptor for rebuilt reactor plugins

### LINKED ISSUE #1555: reactor-plugins are not reloaded on subsequent invocations
### Affected version

1.0.4

### Bug description

I think this is a bug, but could also be seen as a feature request.

## Description

When a project contains a maven-plugin, e.g. https://github.com/mthmulders/polly/pull/877, then subsequent invocations of mvnd will not reload the changed plugins.

For example, let's assume the Mojo is not thread safe.

* change the Mojo to `threadSafe = false`
* run `mvnd verify`
* see the "threadSafe Warning`

Then

* change the Mojo back to `threadSafe = true`
* run `mvnd verify` again
* see that the Mojo is still not treated as threadsafe as it does not get reloaded

## Proposed solution

* `mvnd` could just reload -SNAPSHOT versioned plugins of a reactor (or any plugin, really).

That is easy to implement and is probably working for >> 90% of the cases

## Feature vs Bug

I think it is a bug because on `mvn` invocation the plugin is always reloaded and `mvnd` violates the principle of least surprise (from a UX perspective)

HarshMehta112:
## Problem
  
  When a project builds a Maven plugin in its own reactor (e.g. `mvnd verify`),
  subsequent daemon invocations keep serving a **stale plugin descriptor** even
  after the plugin's code changes. The reported symptom is the `threadSafe` flag
  not being re-read: toggle a Mojo's `@Mojo(threadSafe = ...)`, rebuild, and the
  daemon still uses the old value (#877).

  ## Root cause

  `InvalidatingRealmCacheEventSpy` already evicts reactor **plugin realms** and
  **extension realms** at the end of each build (any realm whose URLs point under
  `multiModuleProjectDirectory`). However, the `PluginDescriptor` — which carries
  the `threadSafe` flag, `@Parameter` default values, etc. — is stored in a
  separate singleton cache (`InvalidatingPluginDescriptorCache`) that was never
  evicted there.

  The timestamp-based invalidation cannot cover this case: with `verify`/`test`
  (no `install`), a reactor plugin is resolved to a `target/classes` **directory**.
  A directory's `lastModifiedTime` does not change when a `.class` file inside it
  is rewritten, so the cached descriptor's file state looks unchanged and the stale
  descriptor is reused. (The existing `ModuleAndPluginNativeIT` does not hit this
  because it uses `clean install`, which rewrites the jar in the local repo and
  bumps its mtime, invalidating everything.)

  ## Fix
  
  Extend `InvalidatingRealmCacheEventSpy` to also evict the
  `InvalidatingPluginDescriptorCache` and `InvalidatingPluginArtifactsCache`
  entries that refer to artifacts in the build tree (or match the configured
  `mvnd.pluginRealmEvictPattern`) at `MavenExecutionResult`. Eviction is driven by
  `CacheRecord.getDependencyPaths()`, consistent with the existing realm eviction.

  ## Test
  
  Adds `ReactorPluginDescriptorReloadTest`, which runs `clean package` (no install,
  so the plugin stays a `target/classes` directory), changes a `@Parameter`
  `defaultValue` between two daemon invocations, and asserts the new value takes
  effect. This fails before the fix and passes after.

  > Note: `integration-tests` test sources currently do not compile on master
  > (`JvmTestClient`/`MvndTestExtension` were not updated after the #970 client API
  > refactor), so this IT could not be run locally. The daemon change itself
  > compiles and passes spotless.

  Fixes #1555

### DIFF
--- a/integration-tests/src/test/java/org/mvndaemon/mvnd/it/ReactorPluginDescriptorReloadTest.java
+++ b/integration-tests/src/test/java/org/mvndaemon/mvnd/it/ReactorPluginDescriptorReloadTest.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.mvndaemon.mvnd.it;
+
+import javax.inject.Inject;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.mvndaemon.mvnd.assertj.TestClientOutput;
+import org.mvndaemon.mvnd.client.Client;
+import org.mvndaemon.mvnd.client.DaemonParameters;
+import org.mvndaemon.mvnd.junit.MvndNativeTest;
+import org.mvndaemon.mvnd.junit.TestUtils;
+
+/**
+ * Reproduces the case where a reactor plugin is rebuilt between two daemon invocations without being installed
+ * (so it is resolved as a {@code target/classes} directory). The plugin descriptor (e.g. parameter default values,
+ * the {@code threadSafe} flag, ...) is cached independently from the plugin realm and must be evicted as well,
+ * otherwise the daemon keeps serving the stale descriptor. See
+ * <a href="https://github.com/apache/maven-mvnd/issues/877">#877</a>.
+ */
+@MvndNativeTest(projectDir = "src/test/projects/module-and-plugin")
+class ReactorPluginDescriptorReloadTest {
+
+    @Inject
+    Client client;
+
+    @Inject
+    DaemonParameters parameters;
+
+    @Test
+    void reloadDescriptorOnReactorPluginChange() throws IOException, InterruptedException {
+        /* Note: we run "package" (not "install"), so the plugin is consumed from the reactor as a
+         * target/classes directory rather than from a freshly installed jar in the local repository. */
+        final Path helloPropertyPath =
+                parameters.multiModuleProjectDirectory().resolve("hello/target/hello.property.txt");
+
+        /* Build #1: the "property" parameter has its default value "Hello" */
+        {
+            final TestClientOutput output = new TestClientOutput();
+            client.execute(output, "clean", "package", "-e").assertSuccess();
+            Assertions.assertThat(helloPropertyPath)
+                    .usingCharset(StandardCharsets.UTF_8)
+                    .hasContent("Hello");
+        }
+
+        /* Build #2: change the parameter default value (a plugin descriptor change that does not alter the
+         * mojo's byte code path) and verify the daemon picks it up instead of reusing the stale descriptor. */
+        {
+            final Path mojoPath = parameters
+                    .multiModuleProjectDirectory()
+                    .resolve("plugin/src/main/java/org/mvndaemon/mvnd/test/module/plugin/mojo/HelloMojo.java");
+            TestUtils.replace(mojoPath, "defaultValue = \"Hello\"", "defaultValue = \"Goodbye\"");
+
+            Thread.sleep(200); // avoid intermittent timestamp resolution issues on some filesystems
+
+            final TestClientOutput output = new TestClientOutput();
+            client.execute(output, "clean", "package", "-e").assertSuccess();
+            Assertions.assertThat(helloPropertyPath)
+                    .usingCharset(StandardCharsets.UTF_8)
+                    .hasContent("Goodbye");
+        }
+    }
+}
--- a/daemon/src/main/java/org/mvndaemon/mvnd/cache/invalidating/InvalidatingRealmCacheEventSpy.java
+++ b/daemon/src/main/java/org/mvndaemon/mvnd/cache/invalidating/InvalidatingRealmCacheEventSpy.java
@@ -37,6 +37,7 @@ import org.apache.maven.execution.MavenExecutionRequest;
 import org.apache.maven.execution.MavenExecutionResult;
 import org.apache.maven.project.MavenProject;
 import org.eclipse.sisu.Typed;
+import org.mvndaemon.mvnd.cache.CacheRecord;
 import org.mvndaemon.mvnd.common.Environment;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -51,6 +52,8 @@ public class InvalidatingRealmCacheEventSpy extends AbstractEventSpy {
     private final InvalidatingPluginRealmCache pluginCache;
     private final InvalidatingExtensionRealmCache extensionCache;
     private final InvalidatingProjectArtifactsCache projectArtifactsCache;
+    private final InvalidatingPluginDescriptorCache pluginDescriptorCache;
+    private final InvalidatingPluginArtifactsCache pluginArtifactsCache;
     private Path multiModuleProjectDirectory;
     private String pattern;
     private PathMatcher matcher;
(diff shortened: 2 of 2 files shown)


gnodet (commented):
The daemon fix is correct and well-motivated — the `shouldEvict(CacheRecord)` approach using `getDependencyPaths()` is the right abstraction for path-based eviction, and the `removeIf` lambda correctly widens the record types.

**Test naming convention** (medium): The test class uses `@MvndNativeTest` but is named `ReactorPluginDescriptorReloadTest.java`. In this project, all 17 `@MvndNativeTest` classes are named `*NativeIT.java` (picked up by Failsafe), and all 31 `@MvndTest` classes are named `*Test.java` (picked up by Surefire). This naming is 100% consistent. With the current name, Surefire would try to run this test during the `test` phase, but it requires the native binary. It should be renamed to `ReactorPluginDescriptorReloadNativeIT.java` (see `ModuleAndPluginNativeIT.java` in the same package for the pattern).

**Silent exception catch** (low): The `shouldEvict(CacheRecord)` catch block swallows exceptions without logging. The existing catch blocks in this class follow the same pattern, so this isn't a consistency violation, but adding a `LOG.debug()` in the new method would help troubleshooting since it catches broad `Exception` rather than narrow `URISyntaxException`.

gnodet (commented):
Excellent root cause analysis — the PR description clearly explains why the existing `ModuleAndPluginNativeIT` passes (`install` changes jar mtime) while the reported scenario (`verify`/`package`) fails.

However, the primary fix has a critical gap:

### Descriptor cache eviction is ineffective (high)

The new `shouldEvict(CacheRecord)` relies on `record.getDependencyPaths()`, which for `InvalidatingPluginDescriptorCache.Record` calls `descriptor.getArtifacts()`. However, Maven 4's `PluginDescriptor` copy constructor does **not** copy the `artifacts` field — only `groupId`, `artifactId`, `version`, `goalPrefix`, `pluginArtifact` (singular), `mojos`, `dependencies`, etc. are copied. Since `InvalidatingPluginDescriptorCache` stores a clone via this constructor, the cached `Record` always has `getArtifacts() == null`.

The existing `Optional.ofNullable(...)` guard in `getDependencyPaths()` then produces an empty stream → `anyMatch()` returns `false` → the descriptor entry is **never evicted** — defeating the fix.

The plugin artifacts cache eviction works correctly, but evicting only the artifacts cache without the descriptor cache does not fix the stale descriptor bug (old `@Parameter` defaults, `threadSafe` flag, etc. are still served).

**Suggestion:** Consider matching descriptor cache keys against reactor project coordinates using the key parameter (the `removeIf(BiPredicate<K, V>)` already provides the key — the lambda currently ignores it with `(k, r) -> shouldEvict(r)`).

### Integration test cannot verify the fix (medium)

The test is correctly designed, but as noted in the PR description, the integration-tests module does not compile on master (pre-existing issue from the #970 client API refactor). This means the fix cannot be verified by the test.

integration-tests/src/test/java/org/mvndaemon/mvnd/it/ReactorPluginDescriptorReloadTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.mvndaemon.mvnd.junit.TestUtils;
+
+/**
+ * Reproduces the case where a reactor plugin is rebuilt between two daemon invocations without being installed
+ * (so it is resolved as a {@code target/classes} directory). The plugin descriptor (e.g. parameter default values,
+ * the {@code threadSafe} flag, ...) is cached independently from the plugin realm and must be evicted as well,
+ * otherwise the daemon keeps serving the stale descriptor. See
+ * <a href="https://github.com/apache/maven-mvnd/issues/877">#877</a>.
+ */
+@MvndNativeTest(projectDir = "src/test/projects/module-and-plugin")
```
gnodet, line 42: **Naming convention:** All 17 existing `@MvndNativeTest`-annotated classes use the `*NativeIT` suffix (e.g., `ModuleAndPluginNativeIT`). The closely related `ModuleAndPluginNativeIT.java` uses the same `projectDir` and follows this pattern.

With the current `*Test` name, Surefire will pick this up during the `test` phase (wrong phase, likely missing native infrastructure), while Failsafe will skip it entirely in the `native` profile.

Suggested rename: `ReactorPluginDescriptorReloadNativeIT`

A companion `ReactorPluginDescriptorReloadTest` extending it with `@MvndTest` (like `ModuleAndPluginTest` extends `ModuleAndPluginNativeIT`) would complete the pattern.

integration-tests/src/test/java/org/mvndaemon/mvnd/it/ReactorPluginDescriptorReloadTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.mvndaemon.mvnd.client.DaemonParameters;
+import org.mvndaemon.mvnd.junit.MvndNativeTest;
+import org.mvndaemon.mvnd.junit.TestUtils;
+
+/**
+ * Reproduces the case where a reactor plugin is rebuilt between two daemon invocations without being installed
+ * (so it is resolved as a {@code target/classes} directory). The plugin descriptor (e.g. parameter default values,
+ * the {@code threadSafe} flag, ...) is cached independently from the plugin realm and must be evicted as well,
+ * otherwise the daemon keeps serving the stale descriptor. See
+ * <a href="https://github.com/apache/maven-mvnd/issues/877">#877</a>.
```
gnodet, line 40: This links to [#877](https://github.com/apache/maven-mvnd/issues/877) (Jansi native library loading on Synology NAS), but this PR fixes [#1555](https://github.com/apache/maven-mvnd/issues/1555) (reactor plugins not reloaded). The PR title and body correctly reference #1555.

gnodet (commented):
Nice work on this fix! The root cause analysis is excellent — clearly identifying why timestamp-based invalidation fails for `target/classes` directories and why the existing `ModuleAndPluginNativeIT` test (which uses `clean install`) doesn't catch this.

The core daemon fix is correct: evicting the plugin descriptor and artifact caches alongside realm caches addresses the root cause. Constructor injection via `@Inject` is clean and safe with Sisu.

A few items to address:

1. **Test naming convention (medium):** All 17 existing `@MvndNativeTest`-annotated classes use the `*NativeIT` naming convention (e.g., `ModuleAndPluginNativeIT`). The Surefire/Failsafe configuration depends on this: Surefire runs `*Test` files, Failsafe runs `*IT` files under the `native` profile. The class should be renamed to `ReactorPluginDescriptorReloadNativeIT` to be picked up by the correct test runner.

2. **Wrong issue reference (low):** The Javadoc links to [#877](https://github.com/apache/maven-mvnd/issues/877) (Jansi native library loading) instead of [#1555](https://github.com/apache/maven-mvnd/issues/1555) (reactor plugins not reloaded), which is the issue this PR fixes.

3. **Silent exception in `shouldEvict` (low):** The catch block in the new `shouldEvict(CacheRecord)` method swallows a broad `Exception` without logging. The existing `shouldEvict` methods catch the narrow `URISyntaxException`. Consider adding `LOG.debug("Error checking cache entry for eviction, evicting to be safe", e)` for troubleshooting.

gnodet (changes requested):
**AI Review — Changes Requested** ❌

Thank you for investigating this issue — the root cause analysis in the PR description is excellent and clearly explains why `ModuleAndPluginNativeIT` passes (uses `install` which rewrites the jar mtime) while the reported scenario fails.

However, the core eviction mechanism for the plugin descriptor cache does not work as intended:

### Critical Issue

**`shouldEvict(CacheRecord)` will never evict plugin descriptor entries.** The method relies on `record.getDependencyPaths()`, which for `InvalidatingPluginDescriptorCache.Record` calls `descriptor.getArtifacts()`. However, Maven 4's `PluginDescriptor` copy constructor does **not** copy the `artifacts` field — it only copies `groupId`, `artifactId`, `version`, `goalPrefix`, `pluginArtifact`, `mojos`, `dependencies`, etc.

Since `InvalidatingPluginDescriptorCache` stores cloned descriptors (both `put()` and `get()` call `clone()`), the cached `Record` always has `getArtifacts() == null`. The `Optional.ofNullable(null).orElse(emptyList())` guard produces an empty stream, `anyMatch()` returns `false`, and the descriptor is **never evicted**.

The `pluginArtifactsCache` eviction (line 120) works correctly, but that alone doesn't fix the stale descriptor bug (old `@Parameter` defaults, `threadSafe` flag, etc. remain cached).

**Possible fixes:**
- (a) Modify `Record.getDependencyPaths()` to include `descriptor.getPluginArtifact().getFile().toPath()` (since `pluginArtifact` IS copied and its file points to `target/classes` for reactor plugins)
- (b) Match descriptor cache entries against reactor project coordinates using the descriptor's GAV
- (c) Use the cache key's `toString()` which returns `groupId:artifactId:version`

### Other Issues

1. **Test naming convention:** All 17 existing `@MvndNativeTest` classes use `*NativeIT.java` naming (failsafe). This class uses `*Test.java` (surefire), which would run during the wrong Maven phase when no native executable is available. Rename to `ReactorPluginDescriptorReloadNativeIT.java`.

2. **Wrong issue reference in Javadoc:** Links to `#877` (Jansi native library issue) instead of `#1555` (reactor plugins not reloaded).

3. **Broad exception catch without logging:** The `catch (Exception e)` block silently swallows errors. The existing methods catch the narrower `URISyntaxException`. Consider adding `LOG.debug("Error checking cache entry for eviction", e)` for troubleshooting.

PR #1667, opened 2026-07-02
Visual improvements to worker lines with live test-progress suffix (2.x)

ammachado:
## What

Presentation-only reshaping of mvnd's live terminal display, plus the client-side plumbing for a live per-test progress suffix, on `master` (2.x, Maven 4).

### Live now (driven by existing messages)
- A drawn progress bar with percent prefixed to the status line.
- Concise arrow-style worker lines: `> :module goal (execution)`.
- Dimmed `> IDLE` slots for free worker threads.

### Wired but dormant (client-side only)
- New `PROJECT_TEST_PROGRESS` wire message (`Message.ProjectTestProgressEvent`, round-trip serialization).
- `TerminalOutput.appendTestProgress(...)` renders a styled suffix `[Tests: N, Failures: .., Errors: .., Skipped: ..] Class#method`, with failure/error counts in red.

## Follow-up

The daemon-side feed that emits `PROJECT_TEST_PROGRESS` (the surefire `ForkNodeFactory` bridge) is **not** implemented on `master` yet, so the test-progress suffix stays dormant until a follow-up commit lands it. This is a deliberate, spike-gated separation; a `TODO` on the `PROJECT_TEST_PROGRESS` constant records it.

## Verification

- `common` unit tests pass (message round-trip; bar geometry; suffix formatting and styles), RAT clean.
- `MultiModuleTest` integration test passes (the reshaping did not break message-driven client assertions).
- Built and verified with JDK 17.

🤖 Generated with [Cursor](https://cursor.com)

### DIFF
--- a/common/src/main/java/org/mvndaemon/mvnd/common/Message.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/Message.java
@@ -251,17 +251,6 @@ public abstract class Message {
     private static final int UTF_BUFS_BYTE_CNT = UTF_BUFS_CHAR_CNT * 3;
     private static final ThreadLocal<byte[]> BUF_TLS = ThreadLocal.withInitial(() -> new byte[UTF_BUFS_BYTE_CNT]);
 
-    static void writeNullableUTF(DataOutputStream output, String value) throws IOException {
-        output.writeBoolean(value != null);
-        if (value != null) {
-            writeUTF(output, value);
-        }
-    }
-
-    static String readNullableUTF(DataInputStream input) throws IOException {
-        return input.readBoolean() ? readUTF(input) : null;
-    }
-
     static String readUTF(DataInputStream input) throws IOException {
         byte[] byteBuf = BUF_TLS.get();
         int len = input.readInt();
@@ -654,8 +643,8 @@ public abstract class Message {
 
         public static ProjectTestProgressEvent read(DataInputStream input) throws IOException {
             final String projectId = readUTF(input);
-            final String testClass = readNullableUTF(input);
-            final String testMethod = readNullableUTF(input);
+            final String testClass = readUTF(input);
+            final String testMethod = readUTF(input);
             final int completed = input.readInt();
             final int failures = input.readInt();
             final int errors = input.readInt();
@@ -713,8 +702,8 @@ public abstract class Message {
         public void write(DataOutputStream output) throws IOException {
             super.write(output);
             writeUTF(output, projectId);
-            writeNullableUTF(output, testClass);
-            writeNullableUTF(output, testMethod);
+            writeUTF(output, testClass);
+            writeUTF(output, testMethod);
             output.writeInt(completed);
             output.writeInt(failures);
             output.writeInt(errors);
--- a/common/src/main/java/org/mvndaemon/mvnd/common/Message.java
+++ b/common/src/main/java/org/mvndaemon/mvnd/common/Message.java
@@ -159,7 +159,6 @@ public abstract class Message {
             case PROMPT:
             case PROMPT_RESPONSE:
             case DISPLAY:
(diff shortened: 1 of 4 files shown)


gnodet (commented):
Well-structured UI improvement with solid tests. Two design points to discuss before the daemon-side follow-up lands:

1. **Message priority ordering** (medium): `PROJECT_TEST_PROGRESS` is assigned class order 2 in `getClassOrder()`, which is higher priority than `PROJECT_STARTED` (3) and `MOJO_STARTED` (4). Since the daemon's `sendQueue` is a `PriorityBlockingQueue` ordered by `getClassOrder()`, a test progress message could theoretically be dequeued before the corresponding `PROJECT_STARTED`, causing `projects.get(e.getProjectId())` to return null and silently drop the message. Practical risk is low since the daemon emitter is not yet implemented, but consider placing it at order 5 or higher before the follow-up — the fix is a one-line change.

2. **Removed plugin version from display** (medium): The worker line format changed from `prefix:version:mojo` to `prefix:mojo`, removing the version that was deliberately added in commit 2a8d18d1 (PR #748). The groupId fallback when no goalPrefix is set also dropped the groupId. This appears to be an intentional UX tradeoff for conciseness by the same author, but it does lose debugging info for plugin version conflicts. Worth confirming this is the desired direction.

gnodet (commented):
Well-structured presentation improvement PR with good test coverage. The concise arrow-style worker lines, idle-slot rendering, and progress bar are all nicely implemented.

One item to consider:

### Redundant `writeNullableUTF` / `readNullableUTF` (medium)

The new helper methods introduce a second wire-format convention for nullable strings (boolean prefix) when the existing `writeUTF`/`readUTF` already handle null via the `-1` sentinel convention. Every other nullable string in `Message.java` uses `writeUTF`/`readUTF` directly.

The `ProjectTestProgressEvent` serialization can simply use the existing methods:
```java
// read:
final String testClass = readUTF(input);
final String testMethod = readUTF(input);

// write:
writeUTF(output, testClass);
writeUTF(output, testMethod);
```

This avoids maintaining two incompatible null-encoding conventions in the same class and saves 1 byte per nullable field on the wire.

gnodet (commented):
Re-review after rebase (d85c8af6 → c716d57b): no functional changes detected.

The previous finding about **redundant `writeNullableUTF`/`readNullableUTF`** remains unaddressed — the existing `writeUTF`/`readUTF` already handle null via the `-1` sentinel convention. `ProjectTestProgressEvent` can simply use them directly, avoiding a second incompatible wire-format convention in the same class.

Visual improvements and test coverage continue to look good.

ammachado:
Once #1670 is merged, I'll port it to the `master` branch.

gnodet (approved):
Re-review after new commit (786ec649): the `writeNullableUTF`/`readNullableUTF` finding from the previous review has been fully addressed — the redundant helpers are removed and the existing `readUTF`/`writeUTF` null-sentinel convention is used correctly. The null round-trip test covers this path. No new issues introduced.

**Minor note (not blocking):** `PROJECT_TEST_PROGRESS` still sits at class order 2, higher priority than `PROJECT_STARTED` (3) and `MOJO_STARTED` (4). When the daemon-side emitter lands (follow-up PR), a test-progress message could be dequeued before its corresponding `PROJECT_STARTED`. Consider bumping it to order 5+ before that follow-up merges.

gnodet (approved):
Re-review after commit 7bb1417a: both prior findings are now resolved.

1. **Message priority ordering** — `PROJECT_TEST_PROGRESS` moved from class order 2 to 5 (after `MOJO_STARTED`=4), preventing the dequeue race.
2. **Plugin version display** — `pluginGroupId` and `pluginVersion` restored in the worker line, matching master.

No new issues. Looks good! ✅

slachiewicz (approved):

PR #1684, opened 2026-07-10
Fix Maven fail-mode option parsing for -fae

### LINKED ISSUE #1680: Bug: `-fae` flag causes incorrect path resolution in plugin goals
### Affected version

1.0.6

## Summary
When using `mvnd -fae` (intended as shorthand for `--fail-at-end`), the flag is incorrectly parsed as separate flags (`-f`, `-a`, `-e`), causing plugin goals to receive corrupted or incorrect working directory/base path values. This does not occur with Maven's `mvn -fae` or when using `mvnd --fail-at-end`.

## Expected Behavior
`mvnd -fae` should be equivalent to `mvnd --fail-at-end`, and plugins should execute with the correct base paths regardless of which form is used.

## Actual Behavior
When running `mvnd -fae`, plugins receive incorrect directory paths. For example, the `xml-maven-plugin:transform` goal fails with:

```
[ERROR] The directory /path/to/project/ae/docs, 
which is a base directory of a ValidationSet or TransformationSet, does not exist.
```

The spurious `ae/` directory component appears to come from the misinterpreted `-a` and `-e` flags.

## Steps to Reproduce
1. Use any Maven project with a plugin that operates on file paths (e.g., `xml-maven-plugin`)
2. Run: `mvnd -fae clean install`
3. Observe path corruption in plugin execution

## Workaround
Use the long form flag instead:
```bash
mvnd --fail-at-end clean install
```

## Environment
- mvnd: 1.0.6 darwin-aarch64
- Maven: 3.9.16
- Java: 25.0.3 (Temurin)
- OS: macOS 26.5.1, aarch64 (but also experienced on Linux CI runners)

## Additional Context
- `mvn -fae` works correctly with standard Maven
- `mvnd --fail-at-end` works correctly
- This suggests a flag-parsing regression specific to mvnd's handling of compound short flags

The issue is not immediately obvious to users switching from Maven to mvnd, making it a usability/documentation concern at minimum.

goutamadwant:
Fixes #1680

### Summary

This prevents mvnd's `MAVEN_FILE` option detection from treating Maven fail-mode short options (`-fae`, `-ff`, `-fn`) as `-f` with an attached value.

It keeps the existing supported `-f` and `--file` forms covered by tests, while allowing Maven's own parser to handle the fail-mode flags.

### Verification

- Reproduced the regression with a focused `EnvironmentTest` assertion before the fix.
- `./mvnw --batch-mode --no-transfer-progress -pl client -am -Dtest=EnvironmentTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvnw --batch-mode --no-transfer-progress -pl common,client -am test`
- `git diff --check`

### DIFF
--- a/client/src/test/java/org/mvndaemon/mvnd/client/EnvironmentTest.java
+++ b/client/src/test/java/org/mvndaemon/mvnd/client/EnvironmentTest.java
@@ -53,6 +53,26 @@ public class EnvironmentTest {
         assertEquals("", Environment.MAVEN_DEFINE.removeCommandLineOption(list("--define")));
     }
 
+    @Test
+    void mavenFileArguments() {
+        assertEquals("pom.xml", Environment.MAVEN_FILE.removeCommandLineOption(list("-f", "pom.xml")));
+        assertEquals("pom.xml", Environment.MAVEN_FILE.removeCommandLineOption(list("-fpom.xml")));
+        assertEquals("pom.xml", Environment.MAVEN_FILE.removeCommandLineOption(list("-f=pom.xml")));
+        assertEquals("pom.xml", Environment.MAVEN_FILE.removeCommandLineOption(list("--file", "pom.xml")));
+        assertEquals("pom.xml", Environment.MAVEN_FILE.removeCommandLineOption(list("--file=pom.xml")));
+    }
+
+    @Test
+    void mavenFailModeArgumentsAreNotMavenFileArguments() {
+        for (String failMode : Arrays.asList("-fae", "-ff", "-fn")) {
+            List<String> args = list(failMode, "verify");
+
+            Assertions.assertFalse(Environment.MAVEN_FILE.hasCommandLineOption(args));
+            Assertions.assertNull(Environment.MAVEN_FILE.removeCommandLineOption(args));
+            assertEquals(list(failMode, "verify"), args);
+        }
+    }
+
     private List<String> list(String... items) {
         return new ArrayList<>(Arrays.asList(items));
     }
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, well-tested fix for a real parsing bug where Maven fail-mode flags (`-fae`, `-ff`, `-fn`) were incorrectly consumed by `MAVEN_FILE`'s `-f` prefix matching.

The `MAVEN_FILE_SHORT_OPTION_CONFLICTS` set and `matchesCommandLineOption` helper are correctly scoped, and all three call sites in `hasCommandLineOption`/`getCommandLineOption` are consistently updated. The test covers both positive (`-f` forms) and negative (`-fae`, `-ff`, `-fn`) cases.

Verified via git blame that `-f` is the only short option in `Environment` that conflicts with other Maven CLI flags.

Note: No CI checks have run on this PR yet. CI verification is recommended before merge.

gnodet (approved):
Clean, well-tested fix that correctly prevents Maven fail-mode options (`-fae`, `-ff`, `-fn`) from being misinterpreted as the `-f`/`--file` option with an attached value.

The approach is pragmatic: a hardcoded conflict set (`MAVEN_FILE_SHORT_OPTION_CONFLICTS`) is appropriate since Maven's short option set is very stable. Good job including both positive tests (verifying `-f` variants still work) and negative tests (verifying `-fae`/`-ff`/`-fn` are excluded, and that the args list is left unmodified).

slachiewicz (approved):

PR #1710, opened 2026-08-03
Enable Mimir build cache in CI

slachiewicz:
`mvnd-1.x` has no Mimir setup at all today — no `.github/ci-extensions.xml`, no daemon properties, no cache steps — so there is nothing to bump here. This ports what `master` has (see #1709) rather than upgrading a version.

Added:

- `.github/ci-extensions.xml` — `eu.maveniverse.maven.mimir:extension3:0.12.0`
- `.github/ci-mimir-daemon.properties` — JGroups off, no LAN cache sharing
- `Prepare Mimir` + `Handle Mimir caches` steps in both the default and the native job

## Two deliberate differences from master

**1. The extensions file goes to `.mvn/extensions.xml`, not `~/.m2/extensions.xml`.**

`mvnd-1.x` builds with the Maven 3.9.14 wrapper. Upstream is explicit that the user-wide location is Maven 4 only:

> With Maven 3 create project-wide, or with Maven 4-rc-5+ create user-wide `~/.m2/extensions.xml`

Copying to `~/.m2/` as master does would silently do nothing here — the extension would never load and the cache would never be populated. The existing `.mvn/extensions.xml` is an empty `<extensions/>` element, so overwriting it during the build loses nothing, and no plugin in this branch checks for a clean working tree (no `doCheck` on buildnumber-maven-plugin).

**2. The artifact is `extension3`.**

`eu.maveniverse.maven.mimir:extension` was discontinued after 0.7.9; since 0.8.0 `extension3` is the single artifact covering both Maven 3 and Maven 4. Requirements are Maven 3.9+ and Java 17+ — satisfied by 3.9.14 and the Java 22 CI toolchain.

## Cache keys

Uses the `mimir2-` prefix, same as master after #1709. Caches created on the default branch are visible to other branches, so a shared prefix is only safe because both branches would be on the same 0.12.0 cache format — Mimir 0.11.0 changed that format and requires older caches to be dropped.

## Note

This is new CI infrastructure for this branch, so the first run on each OS populates the cache from scratch and shows no speedup; the benefit appears on subsequent runs.

### DIFF
--- a/.github/ci-extensions.xml
+++ b/.github/ci-extensions.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+<extensions>
+  <extension>
+    <groupId>eu.maveniverse.maven.mimir</groupId>
+    <artifactId>extension3</artifactId>
+    <version>0.12.0</version>
+  </extension>
+</extensions>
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Well-implemented port of Mimir build caching to the mvnd-1.x branch.

**Two deliberate differences from master — both correct:**
1. Project-level `.mvn/extensions.xml` instead of `~/.m2/extensions.xml` — required because mvnd-1.x uses Maven 3.9.14, and user-wide extensions.xml is Maven 4+ only.
2. `extension3` artifact instead of `extension` — the old artifactId was discontinued after 0.7.9.

**Configuration looks good:**
- JGroups properly disabled for CI runners
- Cache key structure with fallback keys (`mimir2-{OS}-{job}-{pomHash}` → `mimir2-{OS}-{job}-` → `mimir2-{OS}-`) allows partial cache hits
- `shell: bash` on the Prepare Mimir step handles cross-platform compatibility
- Overwriting the empty `.mvn/extensions.xml` during CI is harmless

**Minor note:** Cache prefix `mimir2-` aligns with the as-yet-unmerged PR #1709 on master. If this merges first, mvnd-1.x caches are simply isolated from master's current `mimir-` prefix — harmless.

PR #1712, opened 2026-08-03
ci: add ARM64 support to stat-compat shim for GraalVM 25

slachiewicz:
### Summary

Fixes the GraalVM 25 `native-image` build failure on Linux ARM64 runners (`ubuntu-24.04-arm`).

#### Problem
GraalVM 25 native builds call `stat()` directly (compiled against `glibc 2.33+` headers). `glibc.redef.aarch64` redefines `stat` to `stat@GLIBC_2.17` for backward compatibility. However:
1. `stat-compat.c` was hardcoded to `_STAT_VER = 1` (specific to `x86_64`). On `aarch64`, `_STAT_VER_LINUX` is `0`.
2. `stat-compat.c` was missing from the ARM64 `Scrt1.o` link step in `early-access.yaml` and `release.yaml`.

This resulted in `undefined reference to stat@GLIBC_2.17` during `native-image` linking.

#### Fix
1. Update `stat-compat.c` to conditionally support `__aarch64__` using `__xstat(0, path, buf)`.
2. Include `stat-compat.c` compilation and link `stat-compat.o` into `Scrt1.o` for ARM64 in `.github/workflows/early-access.yaml`.
3. Synchronize `.github/workflows/release.yaml` to include ARM64 glibc patching, `stat-compat.o`, and version verification steps.

### DIFF
--- a/.github/workflows/early-access.yaml
+++ b/.github/workflows/early-access.yaml
@@ -173,9 +173,12 @@ jobs:
                 objcopy --redefine-syms=client/src/main/resources/glibc/glibc.redef.aarch64 -- "$input" "$output" 2>/dev/null
               done
 
+          : compile stat backward-compat shim for glibc before 2.33
+          gcc -O3 -Os -Wall -Wextra -Werror -c -o client/target/stat-compat.o client/src/main/resources/glibc/stat-compat.c
+
           : patch gcc startfile
           gcc -O3 -Os -Wall -Wextra -Werror -Wconversion -Wsign-conversion -Wcast-qual -pedantic -c -o client/target/dynamic-libc-start.o client/src/main/resources/glibc/dynamic-libc-start.c
-          ld -r /lib/aarch64-linux-gnu/Scrt1.o client/target/dynamic-libc-start.o -o client/target/graalvm-libs-for-glibc-2.17/Scrt1.o
+          ld -r /lib/aarch64-linux-gnu/Scrt1.o client/target/dynamic-libc-start.o client/target/stat-compat.o -o client/target/graalvm-libs-for-glibc-2.17/Scrt1.o
           objcopy --redefine-syms=client/src/main/resources/glibc/glibc.redef.aarch64 client/target/graalvm-libs-for-glibc-2.17/Scrt1.o 2>/dev/null
 
       - name: 'Build native distribution'
--- a/.github/workflows/early-access.yaml
+++ b/.github/workflows/early-access.yaml
@@ -199,7 +199,7 @@ jobs:
         run: |
           ldd client/target/mvnd
           err=0
-          objdump -T client/target/mvnd | grep GLIBC_ | grep -v 'GLIBC_\([01]\|2\.[0-9]\|2\.1[017]\)[^0-9]' || err=$?
+          objdump -T client/target/mvnd | grep GLIBC_ | grep -v 'GLIBC_\([01]\|2\.[0-9]\|2\.1[0-7]\)[^0-9]' || err=$?
           (( err == 1 ))
 
       - name: 'Upload daemon test logs'
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -80,8 +80,8 @@ jobs:
       - name: 'Maven clean'
         run: ./mvnw clean -Dmrm=false -B -ntp -e
 
-      - name: 'Patch GraalVM libs for only requiring glibc 2.12'
-        if: ${{ env.OS == 'linux' }}
+      - name: 'Patch AMD64 GraalVM libs for only requiring glibc 2.12'
+        if: ${{ env.OS == 'linux' && env.ARCH == 'amd64' }}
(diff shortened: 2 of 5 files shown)


gnodet (changes requested):
The stat-compat.c ARM64 fix and test hermeticity changes are well-designed and correct. However, the release.yaml ARM64 steps will fail on the current matrix setup.

**Issue: release.yaml ARM64 matrix uses x86_64 runner**

The matrix defines `os: [ubuntu-24.04, ...]` with `arch: [x64, arm64]` and `runs-on: ${{ matrix.os }}`. For the `ubuntu-24.04 + arm64` combination, this runs on an x86_64 runner where:
- `/lib/aarch64-linux-gnu/Scrt1.o` does not exist
- `gcc` compiles for x86_64, not aarch64

The new `Patch ARM64 GraalVM libs` step (conditioned on `env.ARCH == 'aarch64'`) will fail on this x86_64 runner.

**early-access.yaml handles this correctly** by using `ubuntu-24.04-arm` as a separate matrix entry (actual ARM runner) and detecting architecture from `${{ runner.arch }}` instead of `${{ matrix.arch }}`. The release.yaml should be updated to match that approach.

**Minor (pre-existing, not blocking):** early-access.yaml line 199 ARM64 verification regex uses `[017]` (matches only chars 0, 1, 7) vs release.yaml correctly using `[0-7]` (range 0-7). The early-access.yaml regex should be updated to `[0-7]` for consistency.

slachiewicz:
Updated PR with fixes for the review feedback:

1. **Fixed  ARM64 matrix runner alignment:**
   - Switched  matrix to  and derived  from  (matching ).
   - This ensures the  build runs on an actual ARM64 runner () rather than an x86_64 runner.

2. **Fixed  GLIBC verification regex:**
   - Fixed regex range from  to  in  line 202 to properly cover GLIBC 2.10 through 2.17 inclusive.

gnodet (approved):
Previous finding (ARM64 runner mismatch in release.yaml) fully addressed. The matrix simplification from a cross-product (`os` × `arch` with excludes) to a flat list (`ubuntu-24.04`, `ubuntu-24.04-arm`, `macos-15`, `windows-2025`) cleanly avoids the runner/arch mismatch. The `Set vars` step correctly derives `ARCH=aarch64` from `runner.arch=ARM64`. Also noted a good bug fix in early-access.yaml: the regex `2\.1[017]` (which only matched digits 0, 1, 7 individually) was corrected to `2\.1[0-7]` to cover the full GLIBC range 2.10–2.17. LGTM!

gnodet (approved):
New commit `22e3594b` adds `--enable-native-access=ALL-UNNAMED` as the default for `mvnd.jvmArgs` — sensible default for Java 17+ to prevent native-access warnings that will become hard errors in future JVM releases. The `mvnd.properties` comment update correctly reflects the new default.

Previously reviewed changes (ARM64 shim, matrix simplification, regex fix) unchanged. LGTM!

slachiewicz:
### Investigation into CI `BootstrapPluginTest` failure

The GraalVM 25 ARM64 `stat@GLIBC_2.17` shim and test hermeticity fixes in this PR are verified and complete.

Separately, `BootstrapPluginTest.cleanInstall` hit a race condition:
```
Caused by: java.nio.file.NoSuchFileException: /.../target/project-local-repo/org.reproducer/bootstrap
    at org.apache.maven.plugins.clean.CleanMojo.execute(CleanMojo.java:277)
```

#### Root Cause
In Maven 4, `ReactorReader` provisions a shared session-scoped repository at `${session.rootDirectory}/target/project-local-repo`.
When `mvnd clean install` executes a multi-module project in parallel (via `SmartBuilder` with multiple threads):
1. **Thread 1** (Root project `clean` phase): `maven-clean-plugin` recursively deletes the root `target/` directory (which includes `target/project-local-repo`).
2. **Thread 2** (Submodule `install`/`build` phase): Submodule `bootstrap` finishes building and `ReactorReader` writes/installs artifacts into `${session.rootDirectory}/target/project-local-repo/org.reproducer/bootstrap/`.

Because `project-local-repo` is located inside `target/`, Thread 1 deletes subdirectories mid-traversal while Thread 2 is actively reading or writing into `project-local-repo`, resulting in `java.nio.file.NoSuchFileException`.

#### Proposed Architectural Fixes
- **In test project (`src/test/projects/bootstrap-plugin/pom.xml`):** Exclude `project-local-repo` from `maven-clean-plugin` target deletion during parallel reactor execution.
- **In Maven 4 Core / `ReactorReader`:** Ensure session-scoped `project-local-repo` is managed with proper lifecycle locking or excluded from module-level `clean` tasks during parallel builds.

slachiewicz:
@gnodet could You help here with direction?

gnodet:
> @gnodet could You help here with direction?
> 
> 

See https://lists.apache.org/thread/2m1p3l1x8xsl9booond3rjh2lcrg2hl1

PR #1714, opened 2026-08-04
Build against Maven 4.0.0-rc-6

slachiewicz:
Moves `maven.version` to the current RC, and the wrapper `distributionUrl` with it so the two do not drift apart.

Note: this repo's workflows only trigger on pushes to `master`/`mvnd-1.x` and on `pull_request`, so it could not be pre-verified on a fork branch — CI runs here on the PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,7 @@
     <groovy.version>5.0.8</groovy.version>
     <jansi.version>2.4.1</jansi.version>
     <jline.version>3.30.16</jline.version>
-    <maven.version>4.0.0-rc-5</maven.version>
+    <maven.version>4.0.0-rc-6</maven.version>
     <required-maven.version>3.9.10</required-maven.version>
 
     <!-- Keep in sync with Maven -->
--- a/pom.xml
+++ b/pom.xml
@@ -88,7 +88,7 @@
 
     <!-- Keep in sync with Maven -->
     <maven.resolver.version>2.0.20</maven.resolver.version>
-    <slf4j.version>2.0.17</slf4j.version>
+    <slf4j.version>2.0.18</slf4j.version>
     <sisu.version>1.1.0</sisu.version>
     <maven.plugin-tools.version>3.15.2</maven.plugin-tools.version>
     <version.plexus-utils>4.0.2</version.plexus-utils>
--- a/pom.xml
+++ b/pom.xml
@@ -81,13 +81,12 @@
     <graalvm.version>25.0.3</graalvm.version>
     <graalvm.plugin.version>1.1.7</graalvm.plugin.version>
     <groovy.version>5.0.8</groovy.version>
-    <jansi.version>2.4.1</jansi.version>
-    <jline.version>3.30.16</jline.version>
+    <jline.version>4.3.1</jline.version>
     <maven.version>4.0.0-rc-6</maven.version>
     <required-maven.version>3.9.10</required-maven.version>
 
     <!-- Keep in sync with Maven -->
-    <maven.resolver.version>2.0.20</maven.resolver.version>
+    <maven.resolver.version>2.0.21</maven.resolver.version>
     <slf4j.version>2.0.18</slf4j.version>
     <sisu.version>1.1.0</sisu.version>
     <maven.plugin-tools.version>3.15.2</maven.plugin-tools.version>
@@ -336,7 +335,7 @@
       </dependency>
       <dependency>
         <groupId>org.jline</groupId>
-        <artifactId>jline-terminal-jansi</artifactId>
+        <artifactId>jansi-core</artifactId>
         <version>${jline.version}</version>
       </dependency>
       <dependency>
--- a/.mvn/release-settings.xml
+++ b/.mvn/release-settings.xml
@@ -1,24 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <settings xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.1.0 http://maven.apache.org/xsd/settings-1.1.0.xsd" xmlns="http://maven.apache.org/SETTINGS/1.1.0"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-    <profiles>
-        <profile>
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -83,7 +83,7 @@
     <groovy.version>5.0.8</groovy.version>
     <jansi.version>2.4.1</jansi.version>
     <jline.version>3.30.16</jline.version>
-    <maven.version>4.0.0-rc-5</maven.version>
+    <maven.version>4.0.0-rc-6</maven.version>
```
gnodet, line 85: The `maven.version` bump to rc-6 needs the companion properties below (lines 90-91, marked `<!-- Keep in sync with Maven -->`) updated to match:
- `maven.resolver.version`: 2.0.20 → **2.0.21**
- `slf4j.version`: 2.0.17 → **2.0.18**
- `jline.version`: 3.30.16 → **4.3.1** (major version change — requires artifact name updates elsewhere in the POM)

gnodet (changes requested):
This PR bumps `maven.version` to `4.0.0-rc-6` but does not update the companion dependency properties explicitly marked `<!-- Keep in sync with Maven -->` in `pom.xml`:

| Property | Current (this PR) | Maven rc-6 ships | 
|---|---|---|
| `maven.resolver.version` | 2.0.20 | **2.0.21** |
| `slf4j.version` | 2.0.17 | **2.0.18** |
| `jline.version` | 3.30.16 | **4.3.1** |

The JLine gap is especially significant — Maven rc-6 moves from JLine **3.x to 4.x**, which includes artifact renames (e.g. `jline-terminal-jansi` → `jansi-core`). The current `pom.xml` still declares dependencies on the old JLine 3.x artifact names, so bumping `maven.version` alone risks build failures or runtime classpath conflicts.

Note: PR #1696 by @gnodet already handles the full rc-6 upgrade comprehensively — including all synced dependencies, artifact renames, and provisio assembly changes — and appears to supersede this PR.

slachiewicz:
Thanks — the three synced properties were right and are now all in: `slf4j.version` 2.0.18, `maven.resolver.version` 2.0.21, `jline.version` 4.3.1, plus the `jline-terminal-jansi` → `jansi-core` rename in `dependencyManagement` and the removal of the dead `<jansi.version>` (nothing in the tree references it). Confirmed against the rc-6 distribution itself: `jansi-core-4.3.1.jar` is present, there is no `jline-terminal-jansi` jar, and resolver/slf4j are at 2.0.21/2.0.18.

On #1696 superseding this one — I do not think it does, on three counts:

- it sets `maven.version` to `4.0.0-SNAPSHOT` (plus `apache-snapshots` `<repositories>`/`<pluginRepositories>` blocks), not the released `4.0.0-rc-6`. Its title says "upcoming rc-6" — it was written on 27 July, before rc-6 existed.
- it does not touch `.mvn/wrapper/maven-wrapper.properties`, so the repo would still build *itself* with rc-5 while targeting rc-6.
- it is a draft, `mergeable: false` / `rebaseable: false` against a base from a week ago (it diffs from jline 3.30.13 and resolver 2.0.18, whereas master is now on 3.30.16 and 2.0.20), and it has no CI runs at all.

Happy to be overruled if the intent is to track `4.0.0-SNAPSHOT` rather than pin the release — but in that case the wrapper still needs updating separately.

One note on the remaining red: the GraalVM `Build native distribution` jobs fail on master too (run 30862194274 on 67d0a23d, `undefined reference to stat@GLIBC_2.17` on ubuntu-24.04-arm), so that failure is not from this PR. #1712 is the fix for that one.

gnodet (commented):
Thanks for updating `slf4j.version` to 2.0.18 — that aligns with Maven rc-6.

One remaining item: `maven.resolver.version` is still at 2.0.20, but Maven 4.0.0-rc-6 ships 2.0.21. This property is under the `<!-- Keep in sync with Maven -->` comment, so it should be bumped as well.

Regarding my previous comment about `jline.version` — upon closer review, that property is in the general dependency versions section, not under the "Keep in sync" comment, and mvnd was already using a different JLine version than Maven before this PR. The JLine 3.x→4.x migration is a larger effort tracked separately (PR #1696). Apologies for the noise on that one.

pom.xml
```diff
@@ -81,14 +81,13 @@
...
-    <maven.version>4.0.0-rc-5</maven.version>
+    <jline.version>4.3.1</jline.version>
+    <maven.version>4.0.0-rc-6</maven.version>
     <required-maven.version>3.9.10</required-maven.version>
 
     <!-- Keep in sync with Maven -->
-    <maven.resolver.version>2.0.20</maven.resolver.version>
-    <slf4j.version>2.0.17</slf4j.version>
+    <maven.resolver.version>2.0.21</maven.resolver.version>
+    <slf4j.version>2.0.18</slf4j.version>
```
gnodet, line 90: `maven.resolver.version` is still 2.0.20 — Maven 4.0.0-rc-6 ships resolver 2.0.21. This property is under the `<!-- Keep in sync with Maven -->` comment and should be updated.

```suggestion
        <maven.resolver.version>2.0.21</maven.resolver.version>
```

gnodet (approved):
All previously raised concerns are now addressed:

- ✅ `maven.resolver.version` bumped to 2.0.21 (matching Maven rc-6)
- ✅ `slf4j.version` bumped to 2.0.18 (from previous commit)
- ✅ JLine 3→4 artifact rename (`jline-terminal-jansi` → `jansi-core`) with `jansi.version` property removed
- ✅ Stale staging repository (`maven-2247`) cleaned up

All "Keep in sync with Maven" properties verified correct against the Maven 4.0.0-rc-6 parent POM. LGTM!

gnodet (approved):
New commit `2ef67ca1` removes the bundled `jul-to-slf4j` from the provisio descriptor — correct, since Maven 4.0.0-rc-6 supplies it in its own distribution. Avoids duplicate JARs.

All five commits remain focused and well-scoped: Maven rc-6 bump, dependency alignment (resolver 2.0.21, slf4j 2.0.18), JLine 3→4 rename, stale staging repo cleanup, and now jul-to-slf4j deduplication. LGTM!

PR #1720, opened 2026-08-04
fix: feed real stdin to native mvnd process in NativeTestClient

ammachado:
## Summary
- `InputStreamNativeIT.installPluginAndCount` fails on CI (native GraalVM builds) because `NativeTestClient` spawns the real `mvnd` binary but never writes to or closes its stdin pipe.
- The test's `RequestInput`/`RequestInputAvailable` message interception in `TestClientOutput` only works for the in-process JVM test client; for a native test the daemon talks to the spawned binary's own real OS-level stdin, which the harness never fed, so the mojo always saw 0 bytes available instead of the expected 10.
- Adds `TestClientOutput.nativeStdin()` (defaults to `null`, so no behavior change for any other test) so a native IT can supply bytes that get written to the child process's real stdin and closed (EOF) before the daemon requests them, mirroring how a real `pipe | mvnd ...` invocation behaves.

Confirmed on `apache/maven-mvnd`'s own `master` CI (e.g. run 30862194274, headSha `67d0a23d`) that this failure is pre-existing and unrelated to any particular feature branch — it reproduces identically on unmodified `master`.

## Test plan
- [x] `./mvnw -pl dist -am install -DskipTests` then `./mvnw -pl integration-tests test -Dtest=InputStreamNativeIT` locally — compiles cleanly; fails only with "mvnd executable does not exist" since no GraalVM `native-image` toolchain is available in this environment (expected/inherent local limitation, not a code issue).
- [x] `./mvnw -pl integration-tests test -Dtest=InputStreamTest` (JVM-mode counterpart) — passes, confirming the new `nativeStdin()` default is a no-op for the existing path.
- [x] CI (GraalVM native builds on ubuntu/macos/windows) to confirm `InputStreamNativeIT.installPluginAndCount` now passes (`ubuntu-24.04-arm` build pending the merge of 5f53708).

### DIFF
--- a/integration-tests/src/test/java/org/mvndaemon/mvnd/assertj/TestClientOutput.java
+++ b/integration-tests/src/test/java/org/mvndaemon/mvnd/assertj/TestClientOutput.java
@@ -71,6 +71,17 @@ public class TestClientOutput implements ClientOutput {
         return messages;
     }
 
+    /**
+     * Bytes to feed to the real OS-level stdin of a natively spawned {@code mvnd} process
+     * (see {@code NativeTestClient}), or {@code null} if the test doesn't need to simulate stdin.
+     * The {@link #accept(Message)}-based {@code RequestInput}/{@code RequestInputAvailable}
+     * interception only works for the in-process JVM test client, since for a native test the
+     * daemon talks to the spawned binary's own real stdin, which this JVM never sees as messages.
+     */
+    public byte[] nativeStdin() {
+        return null;
+    }
+
     public void assertContainsMatchingSubsequence(String... patterns) {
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Well-scoped fix for a real CI regression introduced by PR #1648. The approach correctly addresses the fundamental problem that native-mode tests cannot use in-JVM message-based stdin interception — `NativeTestClient` now feeds raw bytes via `Process.getOutputStream()` when a `TestClientOutput` provides `nativeStdin()`. The default-null design ensures zero impact on existing tests. CI failure confirmed as pre-existing on master (all four GraalVM native jobs fail). LGTM!

slachiewicz (approved):

